PR #8, opened 2020-05-09
docs: fix incorrect variable names

pnfcre:
Hi there,

I noticed that `items` is no longer returned from the hook in the latest release, so I changed it to `virtualItems` in the docs. Would be great if you could also update the version of `react-virtual` used in the Codesandbox!

Kind regards,
Hampus Kraft.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,7 @@ function RowVirtualizerFixed() {
 
 ```js
 const {
-  items: [
+  virtualItems: [
     { index, start, size, end, measureRef },
     /* ... */
   ],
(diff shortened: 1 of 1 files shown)


pnfcre:
Sidenote: The default import of `react-virtual` is currently broken for TypeScript users, they have to explicitly import the `index.js` file, since the `types/index.d.ts` file is currently empty (https://twitter.com/pnfcre/status/1259011000646369280). Until new types are added, would it be possible to add `declare module 'react-virtual/index` to that file? Thanks!

tannerlinsley:
Yes! Please open a PR for that?

pnfcre:
Sure thing!

Issue #12, opened 2020-05-09
Why this library isn't written in Typescript?


enesTufekci:
Just kidding :) I am going to use this library but I need types, I can implement them or convert codebase to typescript if you want. What would be your preference? 

Cheers!

tannerlinsley:
For now please feel free to implement the type defs in a PR. I’ll convert it to Typescript on my own later. 👍

johnnyreilly:
Great to see this happening 🤗

When you make the TypeScript move it's worth taking https://github.com/jaredpalmer/tsdx for a spin. It'll make your life easier. Feel free to ping me if you would like assistance when it comes to TypeScript; I'd be happy to help!

PR #13, opened 2020-05-09
chore: implement types

enesTufekci:
Fixes: #12

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -21,7 +21,7 @@ declare function useVirtual<T>(options: {
   estimateSize: (index?: number) => number
   overscan?: number
   horizontal?: boolean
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -1 +1,32 @@
...
+  measureRef: React.RefObject<any>
+}
+
+declare function useVirtual<T>(options: {
+  size: number
+  parentRef: React.RefObject<T>
+  estimateSize: (index?: number) => number
+  overscan?: number
+  horizontal?: boolean
+  scrollToFn?: (offset: number) => number
```
tannerlinsley, line 24: This actually returns void 0

enesTufekci: Fixed!

PR #17, opened 2020-05-10
optimise measurements calculation

Arjun-sna:
Prevent creating a reversed array on every scroll event by memoizing reversed array.


tannerlinsley:
Nice!

PR #48, opened 2020-08-13
perf: don't update on scroll/rect changes unless virtualized range changes

mogelbrod:
Fixes https://github.com/tannerlinsley/react-virtual/issues/46

- Avoid state updates caused by `useScroll()` that doesn't cause the virtualized range to change
- Remove the unnecessary `reversedMeasurements` array and `scrollOffsetPlusOuterSize` variable
- Remove the no longer used `useScroll` hook


src/index.js
```diff
@@ -143,6 +104,36 @@ export function useVirtual({
...
+    element.addEventListener('scroll', handler, {
+      capture: false,
+      passive: true,
+    })
+
+    return () => {
+      element.removeEventListener('scroll', handler)
+    }
+  }, [parentRef, scrollKey])
+  //*/
```
mogelbrod, line 135: For some reason I couldn't get the hook to work with the existing `useScroll()`. Sometimes the `setRange()` callback would only be called on the initial mount and not during scrolling. Any ideas on how to fix this?

```suggestion
  useScroll(parentRef, ({ [scrollKey]: scrollOffset }) => {
    latestRef.current.scrollOffset = scrollOffset
    setRange(prevRange => {
      console.log('setRange') // FIXME: Sometimes never triggers after the initial mount
      return calculateRange(latestRef.current, prevRange)
    })
  })
```

mogelbrod:
> Looks good! Can you rebase with the last 2 PRs that came in?

Done!

I also removed the `useScroll` hook that [I couldn't get to work](https://github.com/tannerlinsley/react-virtual/pull/48#discussion_r470160422) in favour of some (hopefully) more streamlined code.

src/index.js
```diff
@@ -52,69 +43,56 @@ export function useVirtual({
...
-  }
+    element.addEventListener('scroll', onScroll, {
+      capture: false,
+      passive: true,
+    })
+
+    return () => {
+      element.removeEventListener('scroll', onScroll)
+    }
+  }, [parentRef, scrollKey, size /* required */])
```
quanbrew, line 90: I tried this PR, list doesn't corrently rendering. fixed by 
```suggestion
  }, [parentRef.current, scrollKey, size /* required */])
```

mogelbrod: Good catch! The old `useScroll` hook had some code that handled the parentRef changing which I had omitted. Looks like this simple change should solve that 👍

mogelbrod:
@tannerlinsley This PR should be ready for another pass now 👍

tannerlinsley:
Looks good!

gausie:
@mogelbrod @tannerlinsley I think this change caused https://github.com/tannerlinsley/react-virtual/issues/68 and https://github.com/tannerlinsley/react-virtual/issues/56 because it attempts to interact with the parentRef element without checking whether it has been assigned (e.g. https://github.com/tannerlinsley/react-virtual/blob/master/src/index.js#L83)

PR #53, opened 2020-08-26
fix: return empty list when size is 0

### LINKED ISSUE #49: virtualRow.index is sometimes NaN
I can't seem to reproduce this consistently, but sometimes virtualRow.index is NaN, which causes my application to crash...

### LINKED ISSUE #52: virtualRow will contain one empty item when size is 0 (since v2.2.4)
# Bug Report

I'm experiencing a bug, where when having an empty list (ie. `size` parameter in `useVirtual` is 0), the returned `virtualRow` array still contains a single element. That single element is "empty", as it is an object that only contains a `measureRef` key, and none of the other keys (`index`, `start`, `end`, `size`).

# Reproduction

I've created a minimal reproduction of the issue at https://codesandbox.io/s/amazing-glade-eo7vw?file=/src/index.js which is a fork of the "Dynamic Rows" example. It starts off with having an empty list, and then after 5 seconds, changes the list to have items. The console log (and the error) shows that we start with an empty `virtualRow`, and then get correct `virtualRows` after 5 seconds.
What should happen instead, is that no console logs are present in the beginning, because the `virtualRows` array should be empty.

# Context

I believe this bug was introduced in version `2.2.4` by https://github.com/tannerlinsley/react-virtual/pull/48. Downgrading to version `2.2.3` "fixes" the issue. But as the changes in that PR are rather substantial, I haven't been able to pinpoint exactly where the problem is. Pinging @mogelbrod 

This might be the same bug as #49.

# Workaround

Current workarounds are:
1. downgrade to version `2.2.3`
2. before mapping through the `virtualRows`, ensure that you have at least one raw `item` (and NOT just one `virtualRow`), eg.
```jsx
<div>
  {items.length > 0 && rowVirtualizer.virtualItems.map((virtualRow) => {
    ...return rows
  })}
</div>
```

mogelbrod:
The `range` state doesn't update in response of changes to `size` until the layout effect is applied. I don't want to update state within the render function, so adding a guard seems like the best approach.

Fixes #52
Might fix #49

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -91,8 +91,9 @@ export function useVirtual({
 
   const virtualItems = React.useMemo(() => {
     const virtualItems = []
+    const end = Math.min(range.end, measurements.length - 1)
 
-    for (let i = range.start; i <= range.end; i++) {
+    for (let i = range.start; i <= end; i++) {
       const measurement = measurements[i]
 
       const item = {


adbayb:
@mogelbrod Is it possible to add a test to avoid future regressions ? Thank you

PR #75, opened 2020-10-22
feat: Add ability to replace useRect with custom observer

o-alexandrov:
This PR should allow users (developers) to greatly increase performance of the tracking measurements of the `parentRef` in modern browsers.
It would also allow to remove `useRect` in production as part of the dead code removal.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,14 +14,14 @@ export function useVirtual({
   parentRef,
   horizontal,
   scrollToFn,
-  observer,
+  useObserver,
 }) {
   const sizeKey = horizontal ? 'width' : 'height'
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -14,12 +14,14 @@ export function useVirtual({
   parentRef,
   horizontal,
   scrollToFn,
+  observer,
```
tannerlinsley, line 17: Since this is technically executed as a custom hook, the name should be `useObserver` or something similar.

tannerlinsley (changes requested):

src/index.js
o-alexandrov: @tannerlinsley Thank you, please review the updated naming

PR #80, opened 2020-11-19
feat: add support for window scrolling

petersendidit:
This adds the ability to use the Window to control the scrolling for a
virtual area rather than that areas scroll. To make this possible two
new options need to be added to `useVirtual`, `onScrollElement` and
`scrollOffsetFn`. These allow the `useVirtualWindow` hook to work
correctly with the existing logic to use Window to scroll the
`parentRef` area.


mogelbrod:
Would you be interested in extending this feature to supporting any closest scroll parent rather than just the window?

I've implemented [a useVirtualScrollParent() hook](https://gist.github.com/mogelbrod/7029848de62016bf0f0b38fb56f8da3e) on top of the current `useVirtual()` which proxies the `parentRef` to enable this, maybe you can find something useful in there. It's being used in production and should work in IE11 along with all the ever green browsers.

petersendidit:
I have opened #95 which only adds the two `onScrollElement` and `scrollOffsetFn` options to `useVirtual`.  I will update this PR when I get the chance to split the `useVirtualWindow` into its own file so it doesn't increase the bundle size.

allocenx:
When will this be merged?!

afzalsayed96:
Hi, Sorry to issue bump this. I just wanted to let you know that this is something I've been waiting for to get merged. And if there's anything I can do to take this PR forward then I'd be happy to help.

popuguytheparrot:
Sorry to issue bump this. What about merge?

mogelbrod:
> When will this be merged?!
> 
> @mogelbrod
> 
> Thank you! ☺️

We've been using the workaround I shared a gist of in production for over a year, but I'm afraid I currently don't have time to submit a PR to integrate this directly into react-virtual myself. It would be amazing if somebody else were to pick it up and convert to a PR though 👍

circlingthesun:
@popuguytheparrot I forked the code and have been using this internally: https://github.com/circlingthesun/react-virtual. It works the same with an element or a window ref.

tannerlinsley:
Let's get this PR up to date with the new docs site and the recent changes to key extraction. Then we'll reassess 👍

thanhlmm:
Hi, do we have any changes to make this merge? Any remaining work

Andrey-Bazhanov:
@thanhlmm we have reimplementation of this request #210, but looks like it also stuck on merging stage

PR #95, opened 2021-01-29
feat: support custom scroll options

petersendidit:
This adds two new options which make it easier to customize the scroll
behavior.

`onScrollElement` which makes it possible to use a different
element to bind the `onScroll` event too.

`scrollOffsetFn` allows a custom function to calculate the offset rather
than using the `parentRef`'s width or height.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,12 @@ const {
 - `paddingEnd: Integer`
   - Defaults to `0`
   - The amount of padding in pixels to add to the end of the virtual list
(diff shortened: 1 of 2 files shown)


allocenx:
Sorry but i've tried it and does not work at all...

const rowVirtualizer = useVirtual({
    size: hasNextPage ? flatPosts.length + 1 : flatPosts.length,
    parentRef,
    overscan: 0,
    estimateSize: React.useCallback(() => 48, []),
    onScrollElement: React.useRef(document.documentElement)
  });

PR #107, opened 2021-03-20
feat: support caching dynamic measurements by unique key

### LINKED ISSUE #83: jank on dynamic infinite lists
I've taken the infinte scrolling example and made it dynamic height based on the index. The list jumps down whenever new rows have been fetched.

https://codesandbox.io/s/eager-williamson-2ydyo?file=/src/index.js

mattmikolay:
This PR introduces a `keyExtractor` option to the `useVirtual` hook, which allows dynamically measured items to be cached by unique ID. This will fix #83.

With these changes, the measure cache is no longer cleared when the list's `size` changes. I've been using these changes for an infinitely scrolling list of dynamically measured elements, and I haven't encountered any issues.

A few notes:
- I'm open to suggestions for a better name than `keyExtractor`. (I took the name `keyExtractor` from [React Native's `FlatList`](https://reactnative.dev/docs/flatlist#keyextractor).)
- There were a few lines in `src/index.js` that Prettier updated automatically.
- The only change I made to `types/index.d.ts` was to add `keyExtractor`…but this file's line endings were CRLF rather than LF. So, EditorConfig automatically converted all line endings to LF when I added my changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,10 @@ const {
  - `scrollOffsetFn: Function() => number`
    - Optional
    - This function, if passed, is called on scroll to get the scroll offest rather than using `parentRef`'s `width` or `height`
+- `keyExtractor: Function(index) => String | Integer`
+  - Optional
+  - This function receives the index of each item and should return the item's unique ID.
+  - This function should be passed whenever dynamic measurement rendering is enabled and the size or order of items in the list changes.
 
 ### Returns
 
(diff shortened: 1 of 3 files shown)


mattmikolay:
A bit hacky, but here's an example of how this works:

https://gist.github.com/mattmikolay/33ac14da335e5a6bef1a8fd6892f6126

PR #111, opened 2021-04-13
scrollOffsetFn not included in effect dependency list

piecyk:
This PR add's 

* update eslint config to include a warning for useIsomorphicLayoutEffect
* add scrollOffsetFn to useIsomorphicLayoutEffect dependency list
* pass event from on scroll to scrollOffsetFn, with this we can easy detect if scrollOffsetFn was called while scroll.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,5 +6,11 @@
   },
   "parserOptions": {
     "sourceType": "module"
+  },
+  "rules": {
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      { "additionalHooks": "(useIsomorphicLayoutEffect)" }
+    ]
   }
 }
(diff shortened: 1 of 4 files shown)


tannerlinsley:
This means the function that the user provides would need to be stable and memoized, which is cumbersome and unnecessary for the user. The function shouldn’t be a part of the dep list itself, so instead, create a ref to track the latest known callback function and use the ref value instead so it doesn’t trigger the effect change detection

piecyk:
> This means the function that the user provides would need to be stable and memoized, which is cumbersome and unnecessary for the user. The function shouldn’t be a part of the dep list itself, so instead, create a ref to track the latest known callback function and use the ref value instead so it doesn’t trigger the effect change detection

Yep, good point 👍

PR #113, opened 2021-04-13
fix: visible range isn't updated when measurements change

### LINKED ISSUE #112: Visible range isn't updated when measurements change
Hi 

noticed that when measurements change after initial render, visible range isn't updated. 
Range is updated on initial render, and next on every scroll.

Maybe let's add isScrolling on state and have an extra effect 

```tsx
  if (!isScrolling && latestRef.current.scrollOffset !== undefined) {
    setRange(prevRange => calculateRange(latestRef.current, prevRange))
  }
}, [isScrolling, measurements])
```

piecyk:
closes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,7 @@ const {
     - `center` places the offset in the center of the visible scroll area
     - `end` places the offset at the bottom/right of the visible scroll area
     - `auto` brings the offset into the visible scroll area either at the start or end, depending on which is closer. If the offset is already in view, it is placed at the `top/left` of the visible scroll area.
+- `isScrolling` - this parameter will be true while the list is being scrolled.
 
 # Contributors ✨
 
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ import React from 'react'
 
 import useRect from './useRect'
 import useIsomorphicLayoutEffect from './useIsomorphicLayoutEffect'
+import { requestTimeout, cancelTimeout } from './timer'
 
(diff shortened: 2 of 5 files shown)


piecyk:
@tannerlinsley 👋 once you have a moment, please take a look on this. Thanks.

About `isScrolling`... I know that it's adds an extra render after scroll, but having it included in lib would be more useful. We can document that in most cases user should add `pointerEvents: none` while scrolling for perfomance reasons. 

```jsx
<div
  ref={parentRef}
  style={{ height, width: '100%', overflow: 'auto', willChange: 'transform', position: 'relative' }}
>
  <div
    style={{
      height: `${rowVirtualizer.totalSize}px`,
      width: '100%',
      pointerEvents: isScrolling ? 'none' : 'auto',
    }}
  >
    {...}
  </div>
</div>
```

piecyk:
Hi @tannerlinsley any updates on this? Sorry to ping you again 🙈  Maybe i can help and sort any question you have or add more tests 🤔 Would really love to get this released, as it's fixes few issues. Thanks 🙏

dariuszsobkowicz (approved):
lgtm

piecyk:
Any news about this? 😉 Would be useful to have some time range when it will be review or what's it holding this back? 

Maybe other contributors also help resolve the issue 🙏

piecyk:
Rebased with master, cc @tannerlinsley 😉 🙏 

Having `isScrolling` would allow in future to optimize overscan only in direction user is scrolling 👍

PR #114, opened 2021-04-14
feat: use binary search to calculate range

piecyk:
This PR changes how the visual range is calculated. Measurements array is always sorted making it a good fit for binary search.


tannerlinsley:
Ooooo! How much impact does this have on the search perf for larger lists?

piecyk:
> Ooooo! How much impact does this have on the search perf for larger lists?

Hmm hard to say, don't have any concrete numbers 🤔

piecyk:
One options is to use https://reactjs.org/docs/profiler.html

Naive setup, 10000 elements in list, doing fast scroll on 400px height

**actualDuration** - time spent rendering the committed update
**baseDuration** - estimated time to render the entire subtree without memoization

average values from 1000 update in development mode 

* current approach  
  - actualDuration: 1.0399449996184558
  - baseDuration: 0.8706299903424224

* binary search 
  - actualDuration: 0.9268499990343116
  - baseDuration: 0.8252650051144883
  
https://gist.github.com/piecyk/b527e80322269ab27222390f2abf3c54

Looks like most of time the values are really close, that's kinda surprising 🤔 Another question how it will look if the rows will be more expensive and overall the whole page, plus react in production mode etc. 🤷‍♂️ 

I'm pretty sure that using binary search would improve overall perfomance, math is saying that also 😉 

**Benchmarking is hard.**

tannerlinsley:
You rock!

PR #122, opened 2021-04-22
feat: add key to VirtualItem

### LINKED ISSUE #120: Include `keyExtractor` value in `virtualItems.item`
It would be useful to include the value returned by `keyExtractor` in each of the items in `virtualItems`.

That way, if your list was stored as a keyed object, it would be efficient to grab the item data.

piecyk:
closes #120

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,8 +285,9 @@ const {
  - `scrollOffsetFn: Function(event?: Event) => number`
    - Optional
    - This function, if passed, is called on scroll to get the scroll offest rather than using `parentRef`'s `width` or `height`
-- `keyExtractor: Function(index) => String | Integer`
+- `keyExtractor: Function(index: Integer) => String | Integer`
   - Optional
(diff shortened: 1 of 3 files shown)


piecyk:
@tannerlinsley we can also push key on measurements. Wondering now in case of dynamic measurements, overall building measurements is expensive maybe better is to cache keys in separate object, they only would need to be re-created when size or keyExtractor change.

tannerlinsley:
I think that makes sense .

piecyk:
Moving `measureRef` into static cache will fix problem with breaking memo as it will not be re-created on range change ( so basic on scroll ) but on other side when element resize now, useVirtual will not be notified. Current is basic a side effect as when range change we re-create measureRef and as it changed react will call it again again. 

But imho @tannerlinsley this is direction we want to go, in case of resizing user should wrap item with ResizeObserver

PR #128, opened 2021-05-07
feat: add isVisible to VirtualItem

piecyk:
This PR adds isVisible property toVirtualItem, making it easy to find what items are in viewport.


piecyk:
@tannerlinsley what do you think about this one? Having isVisible allows for example, implementing scroll sync to visible item between two lists with different content etc.

PR #160, opened 2021-07-05
fix: measureRef infinite loop in firefox while zoom

### LINKED ISSUE #159: Dynamic row height
How to set up a dynamic row height measured by `measureRef`? I know the `size` property should reflect the element's height but if I'm applying it to the row element, its height is always the estimated one, am i missing something?

```
<Row
    key={virtualRow.index}
    ref={el => measureRef(el)}
    style={{ transform: `translate3d(0, ${virtualRow.start}px, 0)`, height: `${virtualRow.size}px`}}
>
...
```

If i'm not setting up the height property on the row, Firefox crashes (seems like it's related to `setMeasuredCache`).

In all examples, the height of the rows is created randomly, what about setting the height by the row's content height?

piecyk:
closes #159

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,7 +4,14 @@ import useRect from './useRect'
 import useIsomorphicLayoutEffect from './useIsomorphicLayoutEffect'
 
 const defaultEstimateSize = () => 50
+
 const defaultKeyExtractor = index => index
+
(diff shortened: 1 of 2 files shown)


NadavShaar:
Please merge ASAP, it's a major bug fix!

PR #164, opened 2021-07-13
feat: add rangeExtractor

### LINKED ISSUE #163: [feature] Allow keeping already rendered items mounted
It would be nice to have an option to keep already mounted items in the list.
This could largely improve scroll performance for components which are expensive to render by not needlessly unmounting already rendered instances.
Of course this only works if the rendered items are properly optimized by eg. `React.memo`.

Here is [a sample CodeSandbox](https://codesandbox.io/s/react-virtual-keep-mounted-yvg9e)

It works by keeping previously rendered virtualItems in the array and updating them when new items arrive.

However, there is an issue if items change their size which could potentially cause to render overlapping items.
I did not test yet, if this can actually become an issue, but if so, it could only be solved if this feature is integrated into react-virtual because then we would have access to the measurement cache and could prevent outdated item positions.

What do you think of such a feature?

------------

PS: This is the typescript code of the implementation:
```ts
import React from 'react';
import { useVirtual, VirtualItem } from 'react-virtual';

/**
 * Merges two arrays while preserving their order.
 * Ordering is determined by the passed getIndex function.
 * If two items have the same index, the item from the first array (primary) is used so duplicates are eliminated.
 * 
 * @example
 * merge([1, 2, 5, 6, 8, 9], [3, 4, 5, 6, 8, 9], x => x);
 * // returns [1, 2, 3, 4, 5, 6, 8, 9];
 */
export function mergeArraysWithOrder<T>(primary: T[], secondary: T[], getIndex: (item: T) => number): T[] {
    let pi = 0;
    let si = 0;
    const result: T[] = [];
    while (true) {
        if (pi >= primary.length) {
            result.push(...secondary.slice(si));
            break;
        } else if (si >= secondary.length) {
            result.push(...primary.slice(pi));
            break;
        }
        const pIndex = getIndex(primary[pi]);
        const sIndex = getIndex(secondary[si]);
        if (pIndex < sIndex) {
            result.push(primary[pi]);
            pi++;
        } else if (pIndex > sIndex) {
            result.push(secondary[si]);
            si++;
        } else {
            result.push(primary[pi]);
            pi++;
            si++;
        }
    }
    return result;
}

/**
 * Same as useVirtual, but keeps once rendered items mounted
 */
export function useVirtualKeepMounted(options: Parameters<typeof useVirtual>[0]) {
    const result = useVirtual(options);

    const totalCount = options.size;

    const lastVirtualItemsRef = React.useRef<VirtualItem[]>([]);
    const virtualItems = React.useMemo(() => {
        let lastItems = lastVirtualItemsRef.current;
        if (lastItems.length > 0 && lastItems[lastItems.length - 1].index >= totalCount) {
            lastItems = lastItems.filter(x => x.index < totalCount);
        }
        const vItems = mergeArraysWithOrder(
            result.virtualItems, lastItems, x => x.index
        );
        lastVirtualItemsRef.current = vItems;
        return vItems;
    }, [totalCount, result.virtualItems]);

    return { ...result, virtualItems };
}
```

piecyk:
closes #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,7 +289,11 @@ const {
   - Optional
   - This function receives the index of each item and should return the item's unique ID.
   - This function should be passed whenever dynamic measurement rendering is enabled and the size or order of items in the list changes.
-
+- `rangeExtractor: Function(range: {start: number; end: number; overscan: number; size: number}) => number[]`
+  - Optional
+  - Defaults to `defaultRangeExtractor`, is exported
+  - **Must be memoized using `React.useCallback()`**
+  - This function receives visible range parameters and should return array of indexes to render
 ### Returns
 
 - `virtualItems: Array<item>`
--- a/src/index.js
+++ b/src/index.js
@@ -13,6 +13,19 @@ const defaultMeasureSize = (el, horizontal) => {
   return el[key]
 }
 
+export const defaultRangeExtractor = range => {
+  const start = Math.max(range.start - range.overscan, 0)
+  const end = Math.min(range.end + range.overscan, range.size - 1)
+
+  const arr = []
+
+  for (let i = start; i <= end; i++) {
+    arr.push(i)
+  }
+
+  return arr
+}
+
 export function useVirtual({
   size = 0,
   estimateSize = defaultEstimateSize,
@@ -27,6 +40,7 @@ export function useVirtual({
   scrollOffsetFn,
   keyExtractor = defaultKeyExtractor,
   measureSize = defaultMeasureSize,
+  rangeExtractor = defaultRangeExtractor,
 }) {
   const sizeKey = horizontal ? 'width' : 'height'
   const scrollKey = horizontal ? 'scrollLeft' : 'scrollTop'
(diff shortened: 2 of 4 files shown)


tannerlinsley:
Love this. 🎉

tannerlinsley:
Would you be up to adding an example to both the sandboxes and the docs?

piecyk:
> Would you be up to adding an example to both the sandboxes and the docs?

Yes, also would be good to update other example, something is wrong with Infinite Scrolling

jonahallibone:
@piecyk Sort of tangentially related, but I saw on the original issue #163 that an `isVisible` key value was in the works. I am wondering if this value could be used to memoize rows to also prevent rerenders on scroll. While this addition is helpful for performance, I think preventing row rerenders on scroll unnecessarily is probably something i'd appreciate for my infinite scrolling implementation

piecyk:
@jonahallibone you can do it now, by using memo on row. One problem is in dynamic context as the `measureRef` is re-crated for every range change, but you can always have an extra div wrapping the memo item.

Also we can move `measureRef` to measurements then problem that it's breaking memo will be resolved.

PR #165, opened 2021-07-19
fix: range is calculated only on scroll

### LINKED ISSUE #112: Visible range isn't updated when measurements change

piecyk:
Another idea to fix #112

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -69,6 +69,8 @@ export function useVirtual({
     [defaultScrollToFn, resolvedScrollToFn]
   )
 
+  const element = onScrollElement ? onScrollElement.current : parentRef.current
+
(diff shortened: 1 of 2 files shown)


piecyk:
Ideally we should not re-calculate range too often. If measurements change and we are scrolling, and this will do exactly that.

PR #166, opened 2021-07-20
fix: range calculated only on scroll

piecyk:
Yet another fix for #112 😅 Idea here is very simple, when measurements request for new range, but if scroll is happening in between cancel the request.


piecyk:
@tannerlinsley what do you think about this fix? Will not affect perfomance when scrolling, dynamic size tests have now correct rows rendered.

Karpengold:
@tannerlinsley it would be nice to have it in the next release :)

PR #169, opened 2021-08-01
feat: add sticky example

### LINKED ISSUE #115: Any guidance for how to do sticky headers?
vertical-sticky-scroll
I'm looking to implement sticky headers, my naive approach was to do `position: absolute; top: 0;` and put an extra `relative` wrapper around the scrollable area. This however broke the scroll (scroll was locked).

I'm planning to resort to `position: absolute; transform: translateY(${offset}px);` keeping it up to date. I'm wondering if there would be a better way using `position: sticky` or something better yet?

piecyk:
This PR includes a simple example how one could implement list with sticky elements. 

https://codesandbox.io/s/github/piecyk/react-virtual/tree/feat/add-sticky-example/examples/sticky

closes #115

@tannerlinsley what do you think about using vite for examples?


piecyk:
Running at https://codesandbox.io/s/sticky-ttk6y

PR #178, opened 2021-08-06
move options interface definition to separated interface and export it

Karpengold:


types/index.d.ts
```diff
@@ -22,9 +22,7 @@ export interface Range {
   size: number
 }
 
-declare function defaultRangeExtractor(range: Range): number[]
-
-declare function useVirtual<T>(options: {
+export interface ReactVirtualOptions {
```
piecyk, line 25: Missing generic param `T`
```suggestion
export interface ReactVirtualOptions<T> {
```

types/index.d.ts
```diff
@@ -47,7 +45,11 @@ declare function useVirtual<T>(options: {
   onScrollElement?: React.RefObject<HTMLElement>
   scrollOffsetFn?: (event?: Event) => number
   rangeExtractor?: (range: Range) => number[]
-}): {
+}
+
+declare function defaultRangeExtractor(range: Range): number[]
+
+declare function useVirtual<T>(options: ReactVirtualOptions): {
```
piecyk, line 52: Pass generic param to `Options`

```suggestion
declare function useVirtual<T>(options: Options<T>): {
```

types/index.d.ts
piecyk: Also imho would drop the `ReactVirtual` prefix 

```suggestion
export interface Options<T> {
```

User can always rename it when importing, `import { useVirtual, Options as ReactVirtualOptions } from 'react-virtual'`

piecyk (changes requested):

types/index.d.ts
Karpengold: yeah, don't know how I missed generic param :) vs code even highlight it!

types/index.d.ts
Karpengold: updated!

piecyk (approved):

PR #179, opened 2021-08-09
perf: reuse already measure items, include key into virtualItem

### LINKED ISSUE #120: Include `keyExtractor` value in `virtualItems.item`

piecyk:
This PR adds small optimisation when creating measurements, as we don't need to rebuild the list from start, only from where item size changed. In addition added `key` value on virtualItem.

closes #120

### DIFF
--- a/docs/src/pages/docs/api.md
+++ b/docs/src/pages/docs/api.md
@@ -8,7 +8,7 @@ title: API
 ```js
 const {
   virtualItems: [
-    { index, start, size, end, measureRef },
+    { key, index, start, size, end, measureRef },
     /* ... */
   ],
   totalSize,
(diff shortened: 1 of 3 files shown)


piecyk:
@tannerlinsley could you have a look on this when you got a moment? Having a key on item would simplify code in some use-cases 🙏

benG2829:
👋  Is there any update on when / if this PR will be merged?

tannerlinsley:
Sorry for the delay. This is a good addition!

Issue #161, opened 2021-07-06
Infinite scroll example fail with fast scrolling


joccd:
Hi !

Thanks for your amazing work.
In your example : https://codesandbox.io/s/github/tannerlinsley/react-virtual/tree/master/examples/infinite-scroll

If I scroll slowly to bottom -> It works
If I scroll quickly to bottom -> "Loading more..." remains
![image](https://user-images.githubusercontent.com/22839734/124578537-972c2480-de4e-11eb-9cf3-13d3cc93557e.png)

Indeed, lastItem.index and flatPosts.length equal 10

Fail on both Chrome and Firefox, I have a good laptop, rather powerful.

Regards,
Joccd

idfunctor:
@joccd  It's due to a small bug in the example; it's not related to a problem with the library itself.

Here's the updated example. https://codesandbox.io/s/jolly-bartik-t7sxj?file=/src/index.js
Had to change `===` to `>=` (in my example it's line 81)

 @tannerlinsley please feel free to update official example as well.

gregsifr:
@idfunctor Thank you for sharing the solution to the equality bug! Any idea how to make the heights dynamic? I followed the examples but it either errors out or overlaps depending on it I set the value of `height` to `${flatPosts[virtualRow.index]}px` or `${virtualRow.size}px`.

This link should be a minimal example (line 99 and 100 are there those height values are) https://codesandbox.io/s/black-rain-7t80x?file=/src/index.js

PR #194, opened 2021-09-11
Update the height and width values in examples

gregsifr:
The dynamic examples were using heights from the generated data instead of the measurement. That is `${rows[virtualRow.index]}px` vs `${virtualRow.size}px`.

The variable example was doing the opposite (using `${virtualRow.size}px`) even though `ref={virtualRow.measureRef}` was never set in the variable example.

Also update the condition in infinite scroll (#161)

### DIFF
--- a/examples/dynamic/src/index.js
+++ b/examples/dynamic/src/index.js
@@ -75,7 +75,7 @@ function RowVirtualizerDynamic({ rows }) {
                 top: 0,
                 left: 0,
                 width: "100%",
-                height: `${rows[virtualRow.index]}px`,
+                height: `${virtualRow.size}px`,
                 transform: `translateY(${virtualRow.start}px)`
               }}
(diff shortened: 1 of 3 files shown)


TkDodo:
to try out the changes to the examples on this PR, codesandbox preview builds would be ideal, like the ones we have on react-query. I'm happy to PR the config changes, but we'd need @tannerlinsley to enable the bot as far as I remember, given that he's the owner :)

gregsifr:
Updated infinite scroll example:  https://codesandbox.io/s/ecstatic-tharp-u3llm
Updated variable height example: https://codesandbox.io/s/fast-frog-tc4l8?file=/src/index.js
Updated dynamic height example: https://codesandbox.io/s/distracted-tdd-o7wh7?file=/src/index.js

PR #209, opened 2021-10-29
fix: Use `types/index.d.ts` for `types` field

amitdahan:
# Summary

Today, the `types` field is defined as follows:
https://github.com/tannerlinsley/react-virtual/blob/5508e19db343bcf02193a8c977f75f56e3d282cc/package.json#L17

While this ends up working during type-check, this isn't helping the auto-import suggestion in VSCode:

image

(I assume the shortest option from the same package is displayed, having `react-virtual/types` isn't ideal)

This is how it looks like after manually editing `node_modules/react-virtual/package.json` with this PR's changes (and restarting the TS server):

image

---

Additionally, this helps partial matching for autocomplete.

Before:

image

After:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   ],
   "main": "index.js",
   "module": "dist/react-virtual.mjs",
-  "types": "types",
+  "types": "types/index.d.ts",
   "sideEffects": false,
   "scripts": {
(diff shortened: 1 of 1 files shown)


amitdahan:
@tannerlinsley 👋 

Anything I can/should do?

PR #217, opened 2021-11-24
pref: calculate range in every render

piecyk:


piecyk:
Getting measurements and range in same render would allow us to write more predicable scrollToIndex, as having pendingMeasuredCacheIndexesRef we know that we need to defer the scroll to next render, checkout the diff 

```diff
diff --git a/src/index.js b/src/index.js
index 6de4de4..fd9af8d 100644
--- a/src/index.js
+++ b/src/index.js
@@ -261,21 +261,34 @@ export function useVirtual({
     [scrollToOffset, size]
   )
 
+  const scrollToIndexArgsRef = React.useRef(null)
+
   const scrollToIndex = React.useCallback(
     (...args) => {
-      // We do a double request here because of
-      // dynamic sizes which can cause offset shift
-      // and end up in the wrong spot. Unfortunately,
-      // we can't know about those dynamic sizes until
-      // we try and render them. So double down!
+      scrollToIndexArgsRef.current = args
+
       tryScrollToIndex(...args)
-      requestAnimationFrame(() => {
-        tryScrollToIndex(...args)
-      })
     },
     [tryScrollToIndex]
   )
 
+  useIsomorphicLayoutEffect(() => {
+    if (
+      pendingMeasuredCacheIndexesRef.current.length === 0 &&
+      scrollToIndexArgsRef.current
+    ) {
+      const args = scrollToIndexArgsRef.current
+      scrollToIndexArgsRef.current = null
+
+      const [index] = args
+      const isVisible = index >= start && index <= end
+
+      if (!isVisible) {
+        scrollToIndex(...args)
+      }
+    }
+  })
+
   return {
     virtualItems,
     totalSize,
```

Pushing offset to state means that hook will re-render on every scroll event, imho it's not bad as overall this re-renders are fast, if there would be  perfomance issues, correct optimisation is to memo on items. cc @tannerlinsley

PR #219, opened 2021-11-29
feat: add useScroll, extract useElementScroll, useWindowScroll

piecyk:
ready for review, cc @tannerlinsley 

to check it locally 
* yarn start in /react-virtual
* USE_SOURCE=1 yarn start in /react-virtual/examples/window-scroll


Zertz:
I'm not familiar enough to look at the code but the example fits the use case perfectly! 🎉

gitcatrat:
@piecyk Sorry if this is obvious but it's not instantly obvious to me..
Why/when use this instead of `parentRef`? Care to explain in few sentences?

piecyk:
Basic `parentRef` need to be set always, in window mode it's used to find position of the list to correctly adjust offset when scrolling to index, also in dynamic size mode when scrolling up.

gitcatrat:
@piecyk So basically it allows the virtual list to be "among" other elements in window?

I think I can replace this hack then? It added it basically to do just that.

```js
const [topOffset, setTopOffset] = useState(0);

const topOffsetSetter = useCallback((el) => {
  if (el) setTopOffset(el.getBoundingClientRect().top);
}, []);

const virtualizer = useVirtual({
  paddingStart: topOffset,
  // ...
});

return (
  <div ref={topOffsetSetter} className="list" style={{height: virtualizer.totalSize - topOffset}}>
    {/* ... */}
  </div>
);
```

gitcatrat:
@piecyk Helps with dynamic mode huh? I hope so, I'll try it as soon as I can.

I've tried to make `react-virtual` work with images but it has been basically unusable. Still haven't figured it out. If images load in, virtual list items are jumping, especially when scrolling back up, i.e screen capture: [https://we.tl/t-TnZ15zGCWA](https://we.tl/t-TnZ15zGCWA).

bnjmnt4n:
This is a breaking change, since it removed some of the options to the `useVirtual` function. Not sure if React Virtual follows SemVer, but it might be a good practice to do this in a major bump, or possibly make it clearer in this PR and the release notes.

piecyk:
> This is a breaking change, since it removed some of the options to the `useVirtual` function. Not sure if React Virtual follows SemVer, but it might be a good practice to do this in a major bump, or possibly make it clearer in this PR and the release notes.

uh-oh @bnjmnt4n sorry for that, you are right we should bump the major 😞 React Virtual should follow semantic versioning, fyi @tannerlinsley

tannerlinsley:
I’ll take care of this. It will require a revert release and recommit

ISilviu:
Is there any way to specify a ref to the scroll container? The `windowRef` is great, however I have some other elements in the page that are left behind when scrolling the virtual list ...

piecyk:
> Is there any way to specify a ref to the scroll container? The `windowRef` feature is great, however I have some other elements in the page that are left behind when scrolling the virtual list ...

Yes, using the useScroll hook, then you can have custom impl for any scroll container.

ISilviu:
Thank you, is there any example in the docs?

Issue #28, opened 2020-05-19
Resizing rows in dynamic list


bebbi:
I've given the dynamic list of this great lib a spin and encountered two issues:

1. My rows can expand/collapse on click (i.e. change height), but the virtualized list doesn't adapt the item positions. Could this benefit from a `ResizeObserver` hook such as [useMeasure](https://github.com/streamich/react-use/blob/master/docs/useMeasure.md)?
2. It would be very nice to have a mechanism where the dynamic list's overall initial height is set such that it doesn't cut off partial rows. See e.g. the first example shown in [react-collapse](https://www.npmjs.com/package/react-collapse) where the height limit is set by the number of visible paragraphs.

piecyk:
Yes, using `ResizeObserver` is the way to go. Current it's possible to implement this with small row wrapper, checkout demo at 
https://codesandbox.io/s/resizeobserver-react-virtual-6131c?file=/src/index.js

IMHO as ResizeObserver is supported in every modern browser https://caniuse.com/resizeobserver maybe we can officially supporting it.

gitcatrat:
@piecyk Have you implemented resize observer it with image content? I.e image loads after some delay.

Scrolling down seems to work but scrolling up is super janky because old accurate height value is thrown away and replaced by the height again where image is missing/is loading again.

kidroca:
I've found that for any dynamic size to be able to work we shouldn't set a fixed `height` on the row elements 
I also have content that allows to expand an item and it doesn't work unless I substitute `height` for `min-height` 

```js
           {rowVirtualizer.virtualItems.map(virtualRow => (
             <div
               key={virtualRow.index}
               ref={virtualRow.measureRef}
               className={virtualRow.index % 2 ? 'ListItemOdd' : 'ListItemEven'}
               style={{
                 position: 'absolute',
                 top: 0,
                 left: 0,
                 width: '100%',
                 // This change
                 minHeight: `${virtualRow.size}px`,
                 transform: `translateY(${virtualRow.start}px)`,
               }}
             >
               Row {virtualRow.index}
             </div>
           ))}
```

For example my inner content would expand an item from 100px to 350px, but it won't work when the row has a fixed height of 100px 
When I use `min-height` the row is allowed to expand, which in turn gets captured by `measureRef` and the change propagates to `virtualRow.size` and `virtualRow.start` 

I imagine to be able to both expand and shrink we should not be setting any height at all, but only `measureRef`

alvis:
[...]

PR #227, opened 2021-12-04
fix: wrong scroll offset when dynamic mode in window scroll 

piecyk:


piecyk:
Fixed for multiple virtual list, when next getting wrong scroll offset as perv is changing it total size.

piecyk:
@tannerlinsley please have a look, would be good to fix this.

bipinrajbhar:
> @tannerlinsley please have a look, would be good to fix this.

Any updates?

src/index.js
```diff
@@ -84,7 +86,10 @@ export const useElementScroll = ({
   }
 }
 
-const useWindowRect = (windowRef, initialRect = { width: 0, height: 0 }) => {
+export const useWindowRect = (
```
bryan-hunter, line 89: can you add the below to lines 107-108?

```
        width: element.innerWidth ? element.innerWidth : element.offsetWidth,
        height: element.innerHeight ? element.innerHeight : element.offsetHeight,
```

When I add this and your changes in this PR, it allows me to use any outer container, not just Window, because innerWidth/height only exist on window and not other elements

The change makes it some 'outerSize' calculates correctly, so it knows what the start/end indexes are

bryan-hunter: also, thank you for this!  This update came right when I needed it :)

bryan-hunter: I actually ended up having to add one more thing to the `scrollToFn` in `useWindowScroll`:

```
    const scrollElementKey = horizontal ? 'scrollLeft' : 'scrollTop'
    // grab different properties if an element and not 'window'
    const refScrollKey = windowRef.current[scrollKey] === undefined ? scrollElementKey : scrollKey
    const refRectKey = windowRef.current[rectKey] === undefined ? scrollElementKey : rectKey

    const delta = ['ToIndex', 'SizeChanged'].includes(reason) ? windowRef.current[refScrollKey] + parentRef.current.getBoundingClientRect()[rectKey] : 0
    windowRef.current.scrollTo({ [refRectKey]: offset + delta })
```

bryan-hunter: and if anyone else stumbles into this somehow - I had to swap all of the `useLayoutEffect` that were doing `setElement` based on ref.current with just `useEffect`

This was due to using React Suspense - the ref.currents were null on the useLayoutEffect, but somehow weren't on the useEffect - I do not know why, but I think React 18 will solve that - didn't seem like there were any issues or performance differences using useEffect on these either

piecyk: @bryan-hunter thanks, imho we should merge this first, then add new changes. 

This looks interesting 

>  I had to swap all of the useLayoutEffect that were doing setElement based on ref.current with just useEffect

Could you create a codesnabdox with the bug and add new issue.

bryan-hunter: @piecyk 
turns out the useEffect vs useLayoutEffect only helps with a workaround for the real issue (and still doesn't solve the main issue most of the time)

here's a sandbox that reproduces it:

https://codesandbox.io/s/toend-forked-cphpc?file=/src/index.js

Basically, when a component suspends, the fallback component is renderered and that height is 0 due to display 'none' right when it executes.

If the fallback component were exactly the same height as what the eventually rendered component was going to be, I think it'd work, but that's hard to determine especially since this is specifically for dynamic mode and estimateSize is just an estimate and not exact.

Here's the main react issue that talks about it: https://github.com/facebook/react/issues/14536 - so it seems like concurrent mode will solve this, but in the meantime is very tough to troubleshoot and find a workable solution for

piecyk: @bryan-hunter hmm in your example even if post are fetched the component is not rendered that's why list is empty as there are not posts 🤔 

Forked example from Dan https://reactjs.org/docs/concurrent-mode-suspense.html#start-fetching-early, looks ok after adding useVirtual https://codesandbox.io/s/with-suspense-wyt0w 🤷‍♂️

bryan-hunter: @piecyk the example I was going for had a Suspense at the Row level in the virtual list

the example that you're using has suspense wrapping the virtual list

it's only an issue when the row suspends, which is probably the most common use case when using dynamic mode

bryan-hunter: @piecyk this is interesting, I modified your sandbox to next the ProfileDetails per row and made it's fetch time be 2000 ms so it was longer than the posts

https://codesandbox.io/s/with-suspense-forked-3bqcs?file=/src/index.js

when I set the fallback div height to "100%" i get maximum call depth exceeded
if I set the fallback div height to "10px" it renders based on that, and then it updates to real height after scrolling a bit

That sandbox was also using react next (18)

piecyk: @bryan-hunter yeah, update the same stuff 👍 https://codesandbox.io/s/with-suspense-wyt0w but used with 5000, hmm why you need the `fallback div height to "100%"` ? Looking on my version it's not needed. 

> it updates to real height after scrolling a bit

This is expected as hook don't know that row size was changed, it's not tracked automatically, one workaround is to trigger parent to re-render (where the hook is called ) or use something to observe it, like ResizeObserver checkout this #28

bryan-hunter: first off, thanks for responding and being so helpful on everything in this repo :)

I think I am seeing where my thinking was flawed.

I use a fallback skeleton that is 100% width/height for most things like this - and that was assuming the wrapper container was giving it the space it needed

On my Fixed lists, it worked great because the estimateSize was always a real, fixed size so the container wrappers always provided the skeletons the right space to fill to 100%.

the Dynamic lists are actually the opposite - the container can estimate the size and provide it, but ultimately we're asking the library to re-measure it once it renders because the child knows its own height, and that's where the Suspense issue was coming in.

I'm actually wondering...if I just let each row's Suspense carry all the way up above where the measureRef is, maybe it won't execute the measureRef until it's done suspending and has a height.. 🤔 🤔

bryan-hunter: eureka - https://codesandbox.io/s/toend-forked-cphpc?file=/src/index.js

that seemed to work, just had to make sure <Suspense> and the promise were executed above the <div> with the measureRef

bryan-hunter: ended up going with the ResizeObserver - too many other things were suspending on the same page above the list on initial render and no guarantees on display none not messing with the height of a row

I believe I will also need this ResizeObserver on my "container ref" that isn't the windowRef - but again I feel like that is only necessary because of Suspense and the display none.

Thanks for all of the help and making the library so flexible to handle these anomalies - it's even flexible enough to handle React issues :p

bryan-hunter: If you want to check out this Library used heavily in action with a lot of infinite scrolls, check out the RYSE app :p - https://apps.apple.com/us/app/ryse/id1560237752 - it's new and free and wouldn't be nearly as smooth without this nice library

piecyk: @bryan-hunter glad to help, awesome that it worked for you!

tannerlinsley:
We have some conflicts here from reverting the breaking change, so a fixup is necessary. Also, I'm not sure if this PR needs others merged first, or if it needs to be pointed at the `next` branch? @piecyk ?

tannerlinsley:
What exactly are we going to use the next branch for now that we have the Alpha branch?

piecyk:
> What exactly are we going to use the next branch for now that we have the Alpha branch?

So next we should rebased `next` with alpha, and sort things in what direction we want to go 😉  The hook idea for supporting window scroll was not that bad, plus revert changes that triggers re-render on every scroll event.

So to re-cap main idea behind `useScroll` hook we can replace those two

```typescript 
  onScrollElement?: React.RefObject<HTMLElement>
  scrollOffsetFn?: (event?: Event) => number
  ```
 
 and because scroll offset should not trigger re-render we need to pass a callback for it 

```typescript
useScroll?: (
  onScrollRef: React.MutableRefObject<(offset: number) => void>,
  options: {
    parentRef: React.RefObject<T>
    horizontal: boolean
  },
) => {
  scrollToFn: (offset: number) => void
}
```

For simplicity I would only allow to pass custom useScroll when we are not in window scroll, so types for useVirtual could be 

```typescript 
interface BaseOptions<T> {
  size: number
  parentRef: React.RefObject<T>
  estimateSize?: (index: number) => number
  overscan?: number
  horizontal?: boolean
  scrollToFn?: (
    offset: number,
    defaultScrollToFn?: (offset: number) => void,
  ) => void
  paddingStart?: number
  paddingEnd?: number
  initialRect?: Rect
  keyExtractor?: (index: number) => Key
  rangeExtractor?: (range: Range) => number[]
  measureSize?: (el: HTMLElement, horizontal: boolean) => number
}

interface ElementOptions<T> extends BaseOptions<T> {
  useScroll?: (
    onScrollRef: React.MutableRefObject<(offset: number) => void>,
    options: {
      parentRef: React.RefObject<T>
      horizontal: boolean
    },
  ) => {
    scrollToFn: (offset: number) => void
  }
  useObserver?: (ref: React.RefObject<T>, initialRect?: Rect) => Rect
  windowRef?: undefined
}

interface WindowOptions<T> extends BaseOptions<T> {
  useScroll?: undefined
  useObserver?: (ref: React.RefObject<Window>, initialRect?: Rect) => Rect
  windowRef: React.RefObject<Window>
}

export type Options<T> = ElementOptions<T> | WindowOptions<T>
``` 

Then implementation for element scroll can look like

```typescript
export const useElementScroll = <T extends HTMLElement>(
  onScrollRef: React.MutableRefObject<(offset: number) => void>,
  { parentRef,  horizontal }: { parentRef: React.RefObject<T>; horizontal: boolean },
) => {
  const updateCounter = useCounterOnRefChange(parentRef)

  const scrollKey = horizontal ? 'scrollLeft' : 'scrollTop'

    const element = parentRef.current

    if (!element) {
      return
    }

    const handleScroll = () => {
      onScrollRef.current(element[scrollKey])
    }

    handleScroll()

    element.addEventListener('scroll', handleScroll, {
      capture: false,
      passive: true,
    })

    return () => {
      element.removeEventListener('scroll', handleScroll)
    }
  }, [updateCounter, scrollKey, onScrollRef, parentRef])

  const scrollToFn = React.useCallback(
    (offset: number) => {
      if (parentRef.current) {
        parentRef.current[scrollKey] = offset
      }
    },
    [parentRef, scrollKey],
  )

  return {
    scrollToFn,
  }
}
```

Then we can export it, allowing user to re-use that for custom scroll elements 

```typescript 
import { useVirtual, useElementScroll } from '@tanstack/react-virtual'

const parentRef = React.useRef<HTMLDivElement | null>(null) 
const scrollerRef = React.useRef<HTMLDivElement | null>(null)

useVirtual({
  size
  parentRef,
  useScroll: (onScrollRef, options) => useElementScroll(onScrollRef, { ...options, parentRef: scrollerRef })
})

<div>...</div>
  ```

PR #236, opened 2021-12-29
chore: update examples

piecyk:


tannerlinsley:
I am *very* excited for this PR, but it looks like it is assuming that some of the breaking changes we reverted are still there. We'll need to revert those changes to be compatible with the new 2.10.3 release or later. After that, I'm good to merge this ASAP.

piecyk:
@tannerlinsley revert those changes, we are good to go 🚀

PR #244, opened 2022-01-12
fix(types): export all types and interfaces

Dahmon:


tannerlinsley:
This has changes that relied on the breaking change from 10.2. It will need to be fixed up to remove those changes.

tannerlinsley:

PR #266, opened 2022-02-19
feat(scrollOffset): return scrollOffset from useVirtual

dantman:
## What is the change?

useVirtual now returns its scrollOffset state.

Docs, tests, and types for the behaviour have been included.

## Motivation

There are a variety of reasons you may want to implement behaviour that is dependent on the current scroll offset of the virtualizer. e.g. Pagination UI like "displaying rows 1-5" or a "Scroll to top" that only shows when `scrollOffset > tooCloseToStart`.

Right now to implement this kind of behaviour the host component needs to add its own `onScroll` handler and update a state. However `useVirtual` already has a `scrollOffset` state updated once on every scroll event resulting in a re-render. As a result if the host component adds its own `onScroll` that changes state, then the host component will re-render **twice** on every single scroll event. Even though `useVirtual` already has all the information that is needed to implement that kind of behaviour and the 2nd re-render is entirely unnecessary and would halve the component's performance. Additionally the UI could experience tearing as the render where useVirtual's new list of items to display and the render where the pagination UI is updated are separate renders.

Here's a demo showing the double render behaviour which would be avoided if useVirtual didn't hide the scrollOffset.
https://codesandbox.io/s/usevirtual-with-an-extra-scroll-event-3q6jqz?file=/src/main.jsx

### DIFF
--- a/docs/src/pages/docs/api.md
+++ b/docs/src/pages/docs/api.md
@@ -14,6 +14,7 @@ const {
   totalSize,
   scrollToIndex,
   scrollToOffset,
+  scrollOffset,
 } = useVirtual({
   size,
   parentRef,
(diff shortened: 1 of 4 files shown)


duttaoindril:
@tannerlinsley Can we get this merged?

tannerlinsley:

PR #277, opened 2022-04-01
type: Export internal interfaces to avoid ts(2043)

### LINKED ISSUE #276: types: Interfaces are not exported
Under some circumstances, not exporting interfaces can cause Typescript errors:

> Exported variable '[someVariable]' has or is using name 'ScrollToIndexOptions' from external module "./node_modules/react-virtual/types/index" but cannot be named.  `ts(4023)`

Adding `export` to these interfaces solves the problem:

```ts
export interface ScrollToOptions {
  align: ScrollAlignment
}

export interface ScrollToOffsetOptions extends ScrollToOptions {}
export interface ScrollToIndexOptions extends ScrollToOptions {}
```

This change could be forward merged to the `alpha` branch as well.

The other option is changing these interfaces to use `type` instead. If a `type` is used but not exported, Typescript will duplicate instead of naming them as symbols.

If a value uses `useVirtual` and is exported and definition files are required, the `ScrollToOffsetOptions` and `ScrollToIndexOptions` interfaces cannot be resolved and Typescript will throw an error.

## Reproduction sandbox
This Codesandbox demonstrates the error:
https://codesandbox.io/s/react-virtual-export-9syofm?file=/src/App.tsx:118-131

It is possible to work around this error by creating a type to manually map to the return type, but that is not possible with generic factory functions.

NicholasBoll:
Fixes #276

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,11 +1,11 @@
 type ScrollAlignment = 'start' | 'center' | 'end' | 'auto'
 
-interface ScrollToOptions {
+export interface ScrollToOptions {
   align: ScrollAlignment
 }
 
-interface ScrollToOffsetOptions extends ScrollToOptions {}
-interface ScrollToIndexOptions extends ScrollToOptions {}
+export interface ScrollToOffsetOptions extends ScrollToOptions {}
+export interface ScrollToIndexOptions extends ScrollToOptions {}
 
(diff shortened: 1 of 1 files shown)


LynchyC:
@tannerlinsley Is there any chance this can be looked at?

tannerlinsley:

PR #285, opened 2022-04-30
ci: fix branch name

goto-bus-stop:
https://github.com/TanStack/react-virtual/commit/061075751c3fad630710c0f7794e587a6cf9dda0 was not published because CI did not run for it, after the default branch was renamed to `main`.


tannerlinsley:

PR #286, opened 2022-05-03
chore: correct umd development output name

wellyshen:
Correct the output name of UMD development file.


tannerlinsley:

PR #304, opened 2022-06-06
feat: Add Solid JS Virtual Adapter

ardeora:
Hello! 
Hope you had a wonderful weekend. This pull request aims to add the SolidJS Adapter for @tanstack/virtual. Some things to note here. 
- This adapter wraps all `Virtualizer` instances inside a proxy object. This was done to preserve the reactivity of the `getVirtualItems` and `getTotalSize` methods. This makes the virtualizer instance easy to use in `createEffect` functions. And it works with iterative utilities like `For` out of the box.
- I am using `createComputed` and `mergeProps` functions for the Virtualizer `options` object to allow users to pass signals in the constructor options. This helps in dynamically updating properties like `count`, `getScrollElement` functions when they are updated/mounted/unmounted
- One other interesting thing to point out is that, when we use callback refs to run `measureElement` on a virtual Item. The behaviour is a little different from react. This is because Solid`s ref callbacks return dom element refs before they are connected to the DOM and hence we would have to do something like this:
```jsx
<div
  ref={(el) => onMount(() => virtualItem.measureElement(el)}
></div>
```
This doesn't seem too bad. But I can also wrap the `measureElement` function call to run only when the element is mounted . This functionality can be abstracted out behind the adapter but I wanted to ask before working on that.
- I tried to recreate the build config by replicating `react-virtual`'s build config. This might need a closer look. But it built on my machine just fine 😄 

I also have all the examples working (except the smooth scrolling). The links for all the examples are attached below:

Fixed Size Elements -> https://stackblitz.com/edit/vitejs-vite-wq8son?file=src/App.tsx
Variable Size Elements -> https://stackblitz.com/edit/vitejs-vite-eywryr?file=src/App.tsx
Dynamically Sized Elements -> https://stackblitz.com/edit/vitejs-vite-dbs3pm?file=src/App.tsx
Padding Before/After List -> https://stackblitz.com/edit/vitejs-vite-sdtmmt?file=src/App.tsx
Sticky Headers -> https://stackblitz.com/edit/vitejs-vite-2sizk8?file=src/App.tsx
Infinite Async Loading List -> [https://stackblitz.com/edit/vitejs-vite-k2nmiw?file=src/App.tsx](https://stackblitz.com/edit/vitejs-vite-k2nmiw?file=src/App.tsx)

Please let me know if you have any suggested changes and I will try to make those changes ASAP.

Hope you have a wonderful week

### DIFF
--- a/packages/solid-virtual/tsconfig.json
+++ b/packages/solid-virtual/tsconfig.json
@@ -1,0 +1,13 @@
+{
+  "composite": true,
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./build/types"
+  },
+  "files": ["src/index.tsx"],
+  "include": [
+    "src",
+    "../virtual-core/src/utils.ts",
+    "../virtual-core/src/index.tsx"
+  ]
+}
--- a/packages/solid-virtual/tsconfig.json
+++ b/packages/solid-virtual/tsconfig.json
@@ -6,8 +6,6 @@
   },
   "files": ["src/index.tsx"],
   "include": [
-    "src",
-    "../virtual-core/src/utils.ts",
-    "../virtual-core/src/index.tsx"
+    "src"
   ]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ yarn-error.log*
 .history
 size-plugin.json
 stats-hydration.json
-stats-react.json
+stats.json
 stats.html
 .vscode/settings.json
 
--- a/docs/adapters/solid-virtual.md
+++ b/docs/adapters/solid-virtual.md
@@ -1,8 +1,6 @@
 ---
-title: Solid Virtual (Coming Soon)
+title: Solid Virtual
 ---
-
-> ⚠️ This module has not yet been developed. It requires an adapter similiar to `@tanstack/react-virtual` to work. We estimate the amount of code to do this is very minimal, but does require familiarity with the Solid framework. If you would like to contribute this adapter, please open a PR!
 
 The `@tanstack/solid-virtual` adapter is a wrapper around the core virtual logic.
 
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "workspaces": [
     "./packages/virtual-core",
     "./packages/react-virtual",
+    "./packages/solid-virtual",
     "./examples/react/dynamic",
     "./examples/react/fixed",
     "./examples/react/infinite-scroll",
--- a/packages/solid-virtual/package.json
+++ b/packages/solid-virtual/package.json
@@ -1,0 +1,42 @@
+{
+  "name": "@tanstack/solid-virtual",
+  "author": "Tanner Linsley",
+  "version": "3.0.0-beta.1",
+  "description": "Headless UI for virtualizing scrollable elements in Solid",
+  "license": "MIT",
+  "homepage": "https://github.com/tanstack/virtual#readme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tanstack/virtual.git"
+  },
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/"
+  },
+  "keywords": [
+    "react",
+    "vue",
+    "solid",
+    "virtual",
+    "virtual-core",
+    "datagrid"
+  ],
+  "funding": {
+    "type": "github",
+    "url": "https://github.com/sponsors/tannerlinsley"
+  },
+  "module": "build/esm/index.js",
+  "main": "build/cjs/solid-virtual/src/index.js",
+  "browser": "build/umd/index.production.js",
+  "types": "build/types/solid-virtual/src/index.d.ts",
+  "sideEffects": false,
+  "engines": {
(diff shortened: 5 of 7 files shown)


FaberVitale:
I was implementing this feature too but you're implementation looks more stable 👍 

 >     * One other interesting thing to point out is that, when we use callback refs to run `measureElement` on a virtual Item. The behaviour is a little different from react. This is because Solid`s ref callbacks return dom element refs before they are connected to the DOM and hence we would have to do something like this:
> 
> 
> ```js
> <div
>   ref={(el) => onMount(() => virtualItem.measureElement(el)}
> ></div>
> ```
> 
> This doesn't seem too bad. But I can also wrap the `measureElement` function call to run only when the element is mounted . This functionality can be abstracted out behind the adapter but I wanted to ask before working on that.
> 

This is what I was using to work around the ref issue.

```ts
// inside solid-virtual 
export type Consumer<T> = (value: T) => void

export function createItemRef<TItemElement>(
  virtualItem: VirtualItem<TItemElement>,
): Consumer<TItemElement>
export function createItemRef<TItemElement>(
  virtualItemList: VirtualItem<TItemElement>[],
): Consumer<TItemElement>
export function createItemRef<TItemElement>(
  virtualItems: VirtualItem<TItemElement>[] | VirtualItem<TItemElement>,
): Consumer<TItemElement> {
  return (el: TItemElement): void => {
    // SolidJs ref callbacks are invoked before the element is inserted in the DOM.
    // It's imperative that we make measurements after the element is part of the document.
    //
    // We could have used directives but unfortunately there's currently an issue with typescript:
    // - https://github.com/solidjs/solid/issues/569#issuecomment-882721883
    // - https://github.com/solidjs/solid/issues/1005
    onMount(() => {
      const items = Array.isArray(virtualItems) ? virtualItems : [virtualItems]

      for (const item of items) {
        item.measureElement(el)
      }
    })
  }
}
```

FaberVitale:
Are you generating the types in build/types?

I think that you also need to make these changes in `<root>/tsconfig.base.json`

```diff
diff --git a/tsconfig.base.json b/tsconfig.base.json
index e08bba1..95f503a 100644
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -16,7 +16,8 @@
     "baseUrl": ".",
     "paths": {
       "@tanstack/virtual-core": ["packages/virtual-core"],
-      "@tanstack/react-virtual": ["packages/react-virtual"]
+      "@tanstack/react-virtual": ["packages/react-virtual"],
+      "@tanstack/solid-virtual": ["packages/solid-virtual"]
     }
   }
 }
```

FaberVitale:
You can check the WIP changes here: https://github.com/FaberVitale/virtual/tree/feat/add-solid-integration

Note that I've modified also [rollup.config.ts](https://github.com/TanStack/virtual/blob/beta/rollup.config.ts).

You just need to call `yarn run build` from <root>.

tsc -b should generates declaration files

packages/solid-virtual/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "composite": true,
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./build/types"
+  },
+  "files": ["src/index.tsx"],
+  "include": [
+    "src",
+    "../virtual-core/src/utils.ts",
```
tannerlinsley, line 10: These two references should not be necessary.

tannerlinsley (changes requested):

tannerlinsley (approved):

lishine:
Hello,
Please help me.
I am trying Virtualizer to be reactive to data and to actually all the options it receives.
I tried to wrap it in useMemo, like bellow , but then lost the onchage updates. I have tried also wrapping options in createStore - same result.
rows is the actual data.

```ts
	const rowVirtualizer = createMemo(() =>
		createVirtualizer({
			count: rows().length,
			getScrollElement: () => props.parentRef,
			estimateSize: (i = 0) => rows()[i].height + gapY,
			overscan: overscanCount,
			onChange: () => console.log('ONCHANGE'),
		})
	)
```

ardeora:
Hello! you should not wrap createVirtualizer in a a memo call. Solid JS primitives are meant to be instantiated only once. By wrapping the virtualizer in a createMemo, you are creating a new instance of it whenever a signal changes. Try this instead:


```tsx
const rowVirtualizer = createVirtualizer({
    get count() {
        return rows().length
    }
    getScrollElement: () => props.parentRef,
    estimateSize: (i = 0) => rows()[i].height + gapY,
    overscan: overscanCount,
    onChange: () => console.log('ONCHANGE'),
})
```
Using the object getter syntax, we can preserve the reactivity of dependent signals. Here is a reference for more info:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

lishine:
Thank you! I will try it

lishine:
The above change causes the list to not populate when scrolling.
I tried to put it in a codesandbox - only two rows can be seen of a 15 rows.

https://codesandbox.io/s/tender-zeh-rcdp3n?file=/src/virtual-inifinite-list.tsx

PR #324, opened 2022-06-17
Add React v18 strict mode support

filip-kis-fsc:
Hi everyone, thanks for making this awesome library!

The goal of this PR is to add strict mode support for React v18

React v18 in strict mode runs `useEffect` twice in dev mode, as documented in [ReactJS blog](https://reactjs.org/blog/2022/03/29/react-v18.html#new-strict-mode-behaviors)

For this reason, the `cleanup` function is called after the initial mount of the component, which does call/clear `this.unsubs`, but the scrolling element `this.scrollElement` remains set.

When `useEffect` runs the second time, the `_willUpdate` if check
 ```js
if (this.scrollElement !== scrollElement) {...
```
evals as `false`, and prevents new subs from being set, resulting in broken virtualisation behaviour, as noted in issue #322 

This PR aims to solve the issue by enhancing the cleanup function so the aforementioned `if` block evals to `true` on subsequent re-renders so the new subs are properly set.

Here is the relevant `react-virtual` code for context: https://github.com/TanStack/virtual/blob/beta/packages/react-virtual/src/index.tsx#L39-L45

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -284,6 +284,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
   private cleanup = () => {
     this.unsubs.filter(Boolean).forEach((d) => d!())
     this.unsubs = []
+    this.scrollElement = null
   }
 
   _didMount = () => {

### CODE REVIEW

**Review by ardeora — CHANGES_REQUESTED** [review]
I believe `this.scrollElement = null` line addition should be on line 290 over here 
```js
_didMount = () => {
    return () => {
      this.cleanup()
      this.scrollElement = null
    }
  }
```

The reasoning behind this is because `_willUpdate` also performs a cleanup every time the reference for the scroll element changes. With your current suggested changes `_willUpdate` will always unsubscribe and resubscribe to the listeners everytime it runs even when the scrollElement ref remains the same

filip-kis-fsc:
> With your current suggested changes _willUpdate will always unsubscribe and resubscribe to the listeners everytime it runs even when the scrollElement ref remains the same.

I don't understand why this would happen as `_willUpdate` has a guard against unnecessary cleanup (line 298) in case the scroll element didn't change:

https://github.com/TanStack/virtual/blob/f41852b8f011e968eddd0413e43ab8fd5c6c3866/packages/virtual-core/src/index.ts#L295-L301

The `_didMount` cleanup function will run twice:
1. after the initial mount (due to React v18 strict mode)
2. on unmount

https://github.com/TanStack/virtual/blob/f41852b8f011e968eddd0413e43ab8fd5c6c3866/packages/virtual-core/src/index.ts#L289-L293

Other than these two cleanup runs, the `this.scrollElement` reference should be stable and should not lead to `_willUpdate` doing unnecessary unsubscribe/resubscribe runs.

Tav0:
This breaks scrolling FYI. I had to go back to beta 6 to scrolldown properly.

wuarmin:
@Tav0 I cannot reproduce your issue, please create a sandbox. Maybe based on this example:
https://codesandbox.io/s/react-virtual-table-example-7v4vbc?file=/src/main.tsx
Thanks.

PR #349, opened 2022-07-27
Fix: Fix Sticky Example Missing Dependency

ardeora:
This PR adds lodash to the dependencies of the sticky react example. Without this dependency, codesandbox fails to load up. Here is the fixed version using lodash https://codesandbox.io/s/youthful-estrela-vggt56?file=/yarn.lock


piecyk:
Thanks, merged https://github.com/TanStack/virtual/pull/373

PR #350, opened 2022-07-29
feat: Allow explicitly overriding `smoothScroll` in scrollToIndex()/scrollToOffset()

christiankaindl:
From the commit message:

> Allow to explicitly override `smoothScroll` when calling scrollToOffset()/scrollToIndex(). Fall back to the global option `enableSmoothScrolling` when the parameter is not explicitly set.

The default behavior is unchanged. That is, when `enableSmoothScroll` is set to true, both scrollToIndex() and scrollToOffset() use smooth scrolling, and if false they don't.

This makes it possible to set the smooth scrolling behavior at the call-site, which is more flexible. Example usage:
```js
scrollToIndex(10, { smoothScroll: false }) // Explicitly set smooth scroll
scrollToIndex(10, { align: 'center', smoothScroll: false }) // Also explicitly set alignment
scrollToIndex(10) // Falls back to `options.enableSmoothScroll`
```

---

Note: I failed to link the package locally so this PR is not tested yet. I tried running `npm run build` and then `npm link` the react-virtual package so that I can test it locally with my project, but I always ran into errors. I didn't find any contributors guide, so any guidance here would be appreciated :)

### DIFF
--- a/docs/api/virtualizer.md
+++ b/docs/api/virtualizer.md
@@ -227,7 +227,8 @@ Returns the virtual items for the current state of the virtualizer.
 scrollToOffset: (
   toOffset: number,
   options?: {
-    align?: 'start' | 'center' | 'end' | 'auto'
+    align?: 'start' | 'center' | 'end' | 'auto',
+    smoothScroll?: boolean
   }
 ) => void
 ```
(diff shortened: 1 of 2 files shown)


christiankaindl:
If this change is something that you'd consider, the docs would probably also need updating. Should I go ahead and and add that to the PR?

tannerlinsley:
Yep. Docs would be great!

christiankaindl:
Added docs & resolved merge conflicts

piecyk:
Why we have `smoothScroll = true` as default, and added enableSmoothScroll to fallback from it? Should user decide if smoothScroll should be used or not?

PR #353, opened 2022-07-30
Only re-render on range changes

### LINKED ISSUE #352: Excessive re-renders when scrolling
### Describe the bug

Scrolling a virtualized list seems to cause a 2+ re-renders on every scroll event. In the linked Sandbox, up to 90 re-renders are triggered before any items are added or removed from the virtual list.

Possibly related
- https://github.com/TanStack/virtual/discussions/212
- https://github.com/TanStack/virtual/discussions/331

Example of react-virtual 2.8, which does not have this problem: https://codesandbox.io/s/react-virtual-2-8-no-re-renders-xlmned

I would love to help debug the issue, but aren't that experienced with monorepo setups yet. Any guidance is appreciated :)

https://codesandbox.io/s/react-virtual-excessive-re-renders-iqtyuf?file=/src/App.js

### Steps to reproduce

1. Scroll the list continuously (e.g. with a trackpad) or use the button "Scroll 1 row"
2. Observe the console.log's in the console, lots of re-renders are logged

### Expected behavior

I would expect to only see re-renders when items are added or removed from the virtualized list. 

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/15364860/181839191-ef0c4465-fd77-434a-b8a8-6f0eedb84b6d.mov

### Platform

macOS 12.4

Tested on latest Firefox and Safari

### tanstack-virtual version

3.0.0-beta.13

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

christiankaindl:
Only notify clients when the range of items changes, thus avoiding many unnecessary updates. Also improves smooth scrolling performance, which equally triggered massive amounts of re-renders.

The current behavior can still be obtained in user-land by simply attaching a scroll listener to the `scrollElement` and calling `measure()` in the event handler

**Note**: So far only tested with the react-virtual adapter
Fixes #352 

---

_Reduces re-renders from this: (watch the re-renders)_

https://user-images.githubusercontent.com/15364860/181907731-3ab351f9-d706-4066-ac7e-1337af0ed7ff.mov

_To this:_

https://user-images.githubusercontent.com/15364860/181907738-534605ce-6e2f-45d6-9197-4c4961731d26.mov

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -328,14 +328,14 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
       this.unsubs.push(
         this.options.observeElementRect(this, (rect) => {
           this.scrollRect = rect
-          this.notify()
+          this.calculateRange()
         }),
       )
 
       this.unsubs.push(
         this.options.observeElementOffset(this, (offset) => {
           this.scrollOffset = offset
-          this.notify()
+          this.calculateRange()
         }),
       )
     }
@@ -394,6 +394,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
       })
       if (range.startIndex !== this.range.startIndex || range.endIndex !== this.range.endIndex) {
         this.range = range
+        this.notify()
       }
       return this.range
     },
@@ -406,7 +407,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
   private getIndexes = memo(
     () => [
       this.options.rangeExtractor,
-      this.calculateRange(),
+      this.range,
       this.options.overscan,
       this.options.count,
     ],
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -269,6 +269,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
     number,
     (measurableItem: TItemElement | null) => void
   > = {}
+  private range: { startIndex: number, endIndex: number } = { startIndex: 0, endIndex: 0 }
 
   constructor(opts: VirtualizerOptions<TScrollElement, TItemElement>) {
     this.setOptions(opts)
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Love it!

wuarmin:
@christiankaindl thank you for the investigation and solution!

wuarmin:
@christiankaindl @tannerlinsley  I have an issue with the latest **beta.14**. At first render only 1 row (height 21px) in a table (height 500px) is rendered. If I set `overscan` to 5, 6 rows are rendered. If I start scrolling or resize the table-container, the rest of the rows get rendered. Do you know how I can workaround this?

It worked at **beta.13**.  It certainly has something to do with cutting out unnecessary renderings. 

```javascript
// pseudo code

  count: data.length,
  getScrollElement: () => tableContainer.current,
  estimateSize: () => 21,
  overscan: 5
});

<div ref={tableContainer}>
  <table>
    virtualRows
  </table>
</div>
```

At first render `tableContainer.current` is `null`.  This is plain react behavior. I think this is the reason for the issue. WDYT? 
Setting `initialRect` has no effect at the moment.

Thanks

christiankaindl:
Yeah, `initialRect` needs fixing, it doesn't work for me either which would also be nice for server-side rendering.

> If I start scrolling or resize the table-container, the rest of the rows get rendered. Do you know how I can workaround this?

Do you mean the other rows stay visibly blank even after the initial render? If so, could you share a codesandbox with the issue? I can't reproduce that on my end.

I guess the reason for only 1 element being rendered on the first render is that as long as `tableContainer.current` is `null` react-virtual doesn't know the height of the element and thus doesn't know how many items to render. So that is expected behavior.

wuarmin:
@christiankaindl Thanky you for the response. I came to the conclusion, that it has nothing to do with the `null`-container-ref. The reason for the issue is that the data-array-length is 0 at first and get changed from the parent of the virtual-table:

Please checkout following sandbox: https://codesandbox.io/s/react-virtual-only-one-item-is-rendered-at-first-bfoj5r

Your manually useEffect-render does not work around the issue.
Thanks?

christiankaindl:
Ok, I see. It seems there is a bug when dynamically updating `count` it doesn't trigger an update. I quickly skimmed the code but it seemed fine. Could you create a new issue?

wuarmin:
@christiankaindl Thank you. I created an issue!

Issue #346, opened 2022-07-26
InitialOffset seems to have no effect.


wuarmin:
### Describe the bug

I tried to apply `initialOffset` to `useVirtualizer`, but it had no effect. 

### Your minimal, reproducible example

https://codesandbox.io/s/initialoffset-test-lttgyh

### Steps to reproduce

See the sandbox.

### Expected behavior

I expect a scrollOffset.

### How often does this bug happen?

Every time

### Platform

-OS Linux Ubuntu Browser: Firefox 103.0b9 (64-bit)

### tanstack-virtual version

^3.0.0-beta.9

### Additional context

It would be great if you could also define which index is initially visible. Is something like this planned?
I.e.
```javascript
  const rowVirtualizer = useVirtualizer({
    count: data.length,
    initialIndex: { index, options: { align: "center" } },
    getScrollElement: () => tableContainer.current,
    estimateSize: () => 21,
    overscan: 3,
  });
```

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

PR #354, opened 2022-07-30
Fix `initialOffset` not working

christiankaindl:
Explicitly scroll to the correct offset.

The range was initially calculated correctly, but the scroll position of `scrollElement` was 0 (which causes the range to be re-calculated to match the visible area). The same applies for SSR usage: the scroll position cannot be transmitted from the server, so it needs to be restored explicitly on the client.

**Note**: Needs to be tested with a SSR setup
Fixes https://github.com/TanStack/virtual/issues/346

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -330,6 +330,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
       this.cleanup()
 
       this.scrollElement = scrollElement
+      this._scrollToOffset(this.scrollOffset, false)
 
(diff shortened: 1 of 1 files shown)


christiankaindl:
Forgot to attach videos.

*Without the fix (`initialOffset: 300`)*:

https://user-images.githubusercontent.com/15364860/184198229-0b1e01e4-a7b6-4a54-8dff-c073123bfe53.mov


*With the fix (`initialOffset: 300`)*:

https://user-images.githubusercontent.com/15364860/184198309-7292e06f-c028-48d1-8e94-8e717b1c7b95.mov

Issue #359, opened 2022-08-02
`getBoundingClientRect()` is called on every frame, uses significant CPU


christiankaindl:
### Describe the bug

Even after #353 I was still seeing semi-optimal scrolling performance in my app, especially in Safari. So I debugged further and noticed that react-virtual has an always-running `requestAnimationFrame()` loop (even when idle), causing non-trivial CPU usage and lag in some scenarios.

I tracked it down to the @reach/observe-rect dependency which uses a `requestAnimationFrame()` loop which makes an expensive call to `getBoundingClientRect()` on literally every frame:
https://github.com/reach/observe-rect/blob/develop/src/index.ts#L19

### Your minimal, reproducible example

https://33v5sz.sse.codesandbox.io/

### Steps to reproduce

1. Open any example from the docs, e.g. https://33v5sz.sse.codesandbox.io/ (is fullscreen)
2. While idle, capture a performance timeline with the browser dev tools profiler 
3. Observe a constantly running `requestAnimationFrame()` loop, each making an expensive call to `getBoundingClientRect()`

### Expected behavior

There should not be significant CPU usage and no idle requestAnimationFrame()` loop.

### How often does this bug happen?

Every time

### Screenshots or Videos

The following is a performance profile of my app, which has noticeable performance issues because of this, especially in Safari:

Safari performance profile, idle:
<img width="1280" alt="Screen Shot 2022-08-02 at 20 29 51" src="https://user-images.githubusercontent.com/15364860/182481840-07065409-d58c-433b-b198-71a1d9ca710c.png">

You can see the many JS calls because of the `requestAnimationFrame()` loop, as well as over 10% CPU utilization just by idling. Sometimes I have seen up to 45%!

The issue can also be reproduced with the examples in the docs, although not as severe (still up to 5% idle CPU usage):
<img width="1280" alt="Screen Shot 2022-08-02 at 20 38 18" src="https://user-images.githubusercontent.com/15364860/182482378-d1340949-ea09-41e3-8144-7aed194de4aa.png">

### Platform

macOS 12.4, 2020 MacBook Air

Tested in latest Firefox, Chrome and Safari

### tanstack-virtual version

3.0.0-beta.14

### TypeScript version

4.5.0

### Additional context

I'd suggest dropping the dependency on @reach/observe-rect and use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) instead, to stay in sync with the `scrollElement` size. That is the most performant solution.

[...]

PR #360, opened 2022-08-02
feat: Use ResizeObserver to listen to size changes

christiankaindl:
Reduces idle CPU usage and should also improve scrolling performance. The previous approach used a `requestAnimationFrame()` loop with calls to `getBoundingClientRect()` on every frame.

Fixes https://github.com/TanStack/virtual/issues/359

Before
Screen Shot 2022-08-02 at 20 29 51

After (+ dark mode 😄):
Screen Shot 2022-08-03 at 00 54 43

Verified it still works, the number of items increase and decrease when resizing:

https://user-images.githubusercontent.com/15364860/182487611-ec7a2635-e86f-4961-89a5-67908b9641e5.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,10 @@
 {
-  "name": "virtual",
+  "name": "@TanStack/virtual",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
+      "name": "@TanStack/virtual",
       "workspaces": [
         "./packages/virtual-core",
         "./packages/react-virtual",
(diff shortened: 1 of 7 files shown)


tannerlinsley:
Fantastic! Would you mind making this same change in react-charts? ;)

tannerlinsley:
The release is failing because of the dependency removal. It changed the output structure of the rollup files. npm run build locally and ensure the exports in package.json map to the correct output files.

wuarmin:
@tannerlinsley  would it be possible to release a new beta.15?
Thanks!

christiankaindl:
Thanks for merging so fast, awesome! 🙌

> Would you mind making this same change in react-charts? ;)

Sure, I will give it a try, probably later today.

> The release is failing because of the dependency removal. It changed the output structure of the rollup files. npm run build locally and ensure the exports in package.json map to the correct output files.

Should I push to this branch or create a new PR?

Issue #363, opened 2022-08-03
Dynamically updating count seems to not trigger an update


wuarmin:
### Describe the bug

When `dataCount` changes from 0 to 1000, the list is not updated.
```javascript
const rowVirtualizer = useVirtualizer({
  count: dataCount,
  getScrollElement: () => tableContainer.current,
  estimateSize: () => 21,
  overscan: 5
});
```

### Your minimal, reproducible example

https://codesandbox.io/s/react-virtual-only-one-item-is-rendered-at-first-bfoj5r?file=/src/main.tsx

### Steps to reproduce

Checkout following sandbox:
https://codesandbox.io/s/react-virtual-only-one-item-is-rendered-at-first-bfoj5r?file=/src/main.tsx
At first glance only one row will be visible. This is because on first render `dataCount` is 0, but then it changes to 1000, but the list is not updated.

### Expected behavior

All items, which fit in the `tableContainer`, should be rendered.

### How often does this bug happen?

Every time

### Platform

-OS Linux Ubuntu Browser: Firefox 103.0b9 (64-bit)

### tanstack-virtual version

3.0.0-beta.14

### Additional context

#353 

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

Himself65:
Same error here

Himself65:
[v3.0.0-beta.9](https://github.com/TanStack/virtual/releases/tag/v3.0.0-beta.9) runs good

wuarmin:
@christiankaindl Do you have a hint about the way to fix this?

christiankaindl:
Yes, you need to re-calculate the virtual item range in `setOptions()`, which is defined here:
https://github.com/TanStack/virtual/blob/f4698d85f33b4a62ca949c0c2dfa0c05c8ed0f4a/packages/virtual-core/src/index.ts#L286

So call `this.calculateRange()` at the end of `setOptions()`. Tested this locally and it seems to work nicely, will open a PR shortly :)

christiankaindl:
@wuarmin I think this PR should fix your issue -> #366

christiankaindl:
I don't know if that's intentional, but the PR in #366 was not merged yet

SvetaSR:
Hi, any updates for this issue? :( I really don't like that we have to use a workaround.

[...]

PR #366, opened 2022-08-11
fix: Re-calculate range when options change

christiankaindl:
Since https://github.com/TanStack/virtual/pull/353, we don't re-calculate the virtualized range in the render function anymore, but instead only when on scroll or size changes. This introduced a regression, where dynamically updating the `count` or other options did not cause an update to the range, and thus no re-render.

Fix this by explicitly calling `this.calculateRange()` in `this.setOptions()`, effectively makes it reactive to any options changes.

*Toggling the `count` option between 0 and 10000, items update accordingly*:

https://user-images.githubusercontent.com/15364860/184197568-237593ae-e611-49d9-9812-73507da18e8f.mov

Fixes https://github.com/TanStack/virtual/issues/363

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -257,7 +257,7 @@ export interface VirtualizerOptions<
 
 export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
   private unsubs: (void | (() => void))[] = []
-  options!: Required<VirtualizerOptions<TScrollElement, TItemElement>>
+  options!: Required<Omit<VirtualizerOptions<TScrollElement, TItemElement>, 'initialOffset' | 'initialRect'>>
   scrollElement: TScrollElement | null = null
   private measurementsCache: Item[] = []
   private itemMeasurementsCache: Record<Key, number> = {}
@@ -276,11 +276,9 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
   }
 
   constructor(opts: VirtualizerOptions<TScrollElement, TItemElement>) {
+    this.scrollRect = opts.initialRect || { width: 0, height: 0 }
+    this.scrollOffset = opts.initialOffset || 0
     this.setOptions(opts)
-    this.scrollRect = this.options.initialRect
-    this.scrollOffset = this.options.initialOffset
-
-    this.calculateRange()
   }
 
   setOptions = (opts: VirtualizerOptions<TScrollElement, TItemElement>) => {
@@ -290,7 +288,6 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
 
     this.options = {
       debug: false,
-      initialOffset: 0,
       overscan: 1,
       paddingStart: 0,
       paddingEnd: 0,
@@ -302,9 +299,10 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
       enableSmoothScroll: true,
       onChange: () => {},
       measureElement,
-      initialRect: { width: 0, height: 0 },
       ...opts,
     }
+
+    this.calculateRange()
   }
 
   private notify = () => {
--- a/packages/react-virtual/__tests__/index.test.tsx
+++ b/packages/react-virtual/__tests__/index.test.tsx
@@ -125,12 +125,13 @@ test.only('should render given dynamic size after scroll', () => {
     target: { scrollTop: 400 },
   })
 
-  expect(screen.queryByText('Row 3')).not.toBeInTheDocument()
-  expect(screen.queryByText('Row 4')).toBeInTheDocument()
-  expect(screen.queryByText('Row 9')).toBeInTheDocument()
-  expect(screen.queryByText('Row 10')).not.toBeInTheDocument()
+  expect(screen.queryByText('Row 2')).not.toBeInTheDocument()
+  expect(screen.queryByText('Row 3')).toBeInTheDocument()
+  expect(screen.queryByText('Row 6')).toBeInTheDocument()
+  expect(screen.queryByText('Row 7')).not.toBeInTheDocument()
 
-  expect(renderer).toHaveBeenCalledTimes(2)
+  // The component is rendered multiple times, because the range changes after each render when the `estimateSize()` estimate is replaced by the items' actual size
+  expect(renderer).toHaveBeenCalledTimes(5)
 })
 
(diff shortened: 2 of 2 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -276,11 +276,9 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
   }
 
   constructor(opts: VirtualizerOptions<TScrollElement, TItemElement>) {
+    this.scrollRect = opts.initialRect || { width: 0, height: 0 }
```
christiankaindl, line 279: I moved `this.scrollRect` to before the `this.setOptions()` call, because the `this.calculateRange()` call in `setOptions()` otherwise needs an if-guard to prevent a crash on the first init (`this.scrollRect` would not be set on first run).

It being above `this.setOptions()` also means the defaults needed to move to here too. To prevent duplicated defaults (here and in `setOptions()`) I removed `initialRect` and `intialOffset` from the options object altogether, as they are only used in the constructor for initialization anyways. Not sure if that is the best solution, but feel free to object if you'd like to keep `initialRect`/`initialOffset` in the options.

Himself65 (approved):

wuarmin:
@tannerlinsley Can we merge this PR? It would fix [#363](https://github.com/TanStack/virtual/issues/363), which is not fixed yet.
Thanks!

Himself65:
when can this PR land?

zhouzi:
@christiankaindl any idea why the tests are failing? Is it a false positive or should the tests be updated? It'd probably be easier to get it merged if all checks are passing and I am interested in this fix as well 🙏

christiankaindl:
@zhouzi I quickly looked into it: One test doesn't pass and it doesn't seem to be a false positive. I'm not entirely sure what the issue is yet. Will try to do some more debugging in the next days.

For anyone interested, the failing test is:
https://github.com/TanStack/virtual/blob/84663b35d0f71d499ed0d85a0f8f84b4a0391923/packages/react-virtual/__tests__/index.test.tsx#L113

It fails a few lines below at
https://github.com/TanStack/virtual/blob/84663b35d0f71d499ed0d85a0f8f84b4a0391923/packages/react-virtual/__tests__/index.test.tsx#L128
The test scrolls to 400 pixels and then checks if the correct elements are being rendered. I noticed the range is re-calculated correctly initially, but then there are a few extra renders and the range is wrong after that. Don't know why that is happening yet 🤷‍♂️

christiankaindl:
**TL;DR** I think the failing test was not correct. The code that calculates the range was not touched in this PR so I updated the test to match the expected behavior.

---

**Explanation**:
- the container height in the test is 200px
- each item is 100px
- the `estimateSize()` function supplied by the "user" estimates 50px for each item.

In the end the range should only include 2 items (2\*100px=200px) to fill the container fully, but initially it contains 4 items (4\*50px=200px) as react-virtual only knows what `estimateSize()` gives it. When the 4 items are rendered initially, their real sizes are then measured and the cache is updated with those new sizes (100px each). So during the next re-renders the range changes to only include 2 items.
**Note**: it actually always over-renders the range by 2. before it rendered 6 and now it renders 4 items

Without this PR: the end range is `{ startIndex: 5, endIndex: 8 }`, which renders 6 items
With this PR: the end range is `{ startIndex: 4, endIndex: 5 }`, which renders 4 items

@tannerlinsley I updated the tests to check for the new range. Does that work for you, or do you want to investigate further?

christiankaindl:
Now the tests pass but the other check doesn't ^^ Is that related to my changes?

rosskhanas (approved):

Himself65:
anyupdate?

piecyk:
Hi, checked the PR and looks like better if we push this logic to adapters, as we can't know how the options are set there. 

With this change on react adapter it will try to calculateRange on each render, but what we really want to 

```tsx
  const afterFirstEffect = React.useRef(false)

    if (afterFirstEffect.current) {
      instance.calculateRange()
    }
    afterFirstEffect.current = true
  }, [
    options.count,
    options.initialRect,
    options.overscan,
    options.horizontal,
    options.paddingStart,
    options.paddingEnd,
    options.scrollPaddingStart,
    options.scrollPaddingEnd,
    options.initialOffset,
    options.rangeExtractor, // now this we need to wrap
  ])
```
We can also write a function that will do a deep comparison and skip functions, but still would include the rangeExtractor in deps. Overall we should expose calculateRange and range as public api of the Virtualizer

piecyk:
btw, till we fix this, one hack for react adapter can be

```tsx
  count: dataCount,
  estimateSize: () => 50,
});

const virtualizerRef = useLatestRef(virtualizer)

React.useLayoutEffect(() => {
  const v = virtualizerRef.current
  v._didMount()() 
  v._willUpdate()
}, [virtualizerRef, dataCount])
```

francislavoie:
@piecyk where is `useLatestRef` from, in this case? I'd love to use a workaround for now.

piecyk:
@francislavoie yes, basic we don't want to have virtualizer as dep in that effect. It should be well know pattern by now in react community, but if someone don't know it, here is Kent blog post about it https://epicreact.dev/the-latest-ref-pattern-in-react/

christiankaindl:
@piecyk 
> Hi, checked the PR and looks like better if we push this logic to adapters, as we can't know how the options are set there.
> With this change on react adapter it will try to calculateRange on each render, but what we really want in, for example in react adapter

Right, I thought about this as well when I opened this PR. It's a bit more work, but ultimately the more correct/efficient approach. I'll probably be able to make the changes early next week, if the PR is blocked on this.

piecyk:
@christiankaindl on other hand, more robust would be as you did to call the calculateRange when options change, but not for every call of setOptions. Something like https://github.com/TanStack/virtual/compare/beta...piecyk:fix/calculate-range

Here we need to update docs that rangeExtractor and measureElement need to have stale ref as they also affect the range.

christiankaindl:
Yes that would also work and scales better when new framework adapters are added over time. Then the only question is, if having to use `useCallback()` is acceptable API wise (it was like this in v2.x as well)?

piecyk:
> Then the only question is, if having to use useCallback() is acceptable API wise (it was like this in v2.x as well)?

I think it's fine, what do you think @tannerlinsley?

zryambus:
when this PR will be merged?

piecyk:
This issue resolved via https://github.com/TanStack/virtual/pull/414

PR #377, opened 2022-08-21
fix #376: flushing the item ref function and forcing a remeasurement

### LINKED ISSUE #376: Virtualizer does not update after adjusting row height
### Describe the bug

Building on the dynamic example, I added a timeout to update the height of rows. The element size updated but the positions did not. 

https://codesandbox.io/s/delicate-river-3rigv0?file=/src/main.tsx

### Steps to reproduce

Wait for the timeout, see broken table.

### Expected behavior

As the size of rows update, the position of the rows should be adjusted.

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/40192128/185458840-b3020220-63de-49ff-adb9-16e67d63030d.mov

### Platform

I tested on Firefox and Chrome on macOS 12.5

### tanstack-virtual version

3.0.0

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

carlosbaraza:
This solution works. However, this the approach taken to flush the cache is not perfect as the table moves around when running `measure`. The lack of idempotence shows that it could be improved, presumably by maintaining the measurements for non visible items and only recalculating the ones visible.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -600,6 +600,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
 
   measure = () => {
     this.itemMeasurementsCache = {}
+    this.measureElementCache = {}
     this.notify()
   }
 }


carlosbaraza:
@tannerlinsley do you think we could get this one in before final v3?

badsyntax:
This change fixes `.measure()` in my scenario where row element heights have dynamically changed.

piecyk:
This should not be needed with new approach via #405, checkout latest beta!

PR #390, opened 2022-09-12
feat(adapter):  Add Vue Virtual Adapter

s-montigny-desautels:
Hi! I made a Vue adapter for virtual-core.
Here is a stackblitz [demo](https://stackblitz.com/edit/vitejs-vite-kwvjgl?file=src/main.ts). 

I am wrapping the `Virtualizer` instance with a `shallowRef` and manually calling `triggerRef` on it to keep reactivity.
I am also calling both `getScrollElement` and `estimateSize` inside the `watchEffect` since most of the time, an `Element` is inside a `ref` and we need to update the options when it changes.

Plean let me know if you have any suggestions!


Wurielle:
I am using your hook for my project and got it to work, although I encountered a few issues.

My use case is the following: 
I have a custom select component that has a search bar inside it allowing me to filter through options listed in my virtual list. When using the search, my options are filtered which is changing the `count` prop. The `count` property in `opts` under the `watchEffect` contains the correct amount when logged but the amount listed in the dom isn't always correct. I could have a count of 52 options and have only 3 shown in the dom.

Here's what I noticed:
- When changing the count property the dom would sometimes update correctly but not always
- onChange is sometimes triggered when the options change but not always

I tried to make a reproduction example but I also have an issue making the dom to render the proper amount of items.
You can take a look here:
https://codesandbox.io/s/vue-virtual-h9o7p7?file=/src/App.vue

packages/vue-virtual/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      virtualizer.setOptions({
+        ...opts,
+        onChange: (instance: Virtualizer<TScrollElement, TItemElement>) => {
+          // Force an update event
+          triggerRef(state)
+          opts.onChange?.(instance)
+        },
+      })
+
+      virtualizer._willUpdate()
```
Wurielle, line 48: Adding another `triggerRef(state)` at this level fixed the dom update for my use case in my app but does not for the codesandbox example that I provided above

Wurielle (commented):
Adding another triggerRef(state) at this level fixed the dom update for my use case in my app but does not for the codesandbox example that I provided above

packages/vue-virtual/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    })
+
+    onCleanup(virtualizer._didMount())
```
salqueng, line 51: I modified this section to use `pauseTracking` and `resetTracking` from `@vue/reactivity`, and this fixes https://github.com/TanStack/virtual/pull/390#issuecomment-1257154850

```suggestion
    pauseTracking();
    virtualizer.setOptions({
      ...opts,
      onChange: (instance: Virtualizer<TScrollElement, TItemElement>) => {
        // Force an update event
        triggerRef(state)
        opts.onChange?.(instance)
      },
    })

    virtualizer._willUpdate()

    onCleanup(virtualizer._didMount())
    resetTracking();
```

Here is the fork: https://codesandbox.io/s/vue-virtual-forked-p1z06n

s-montigny-desautels: Hi, thank you for that! I did not know those functions existed. They are really convenient. 

I've added that change. It works like a charm now.

salqueng:
@s-montigny-desautels 
After some investigations, I found a way not to use pauseTracking and resetTracking, which are internal-like methods of vue - create a function and call it and use watch.
Also, I found a bug that a virtualizer cannot render initial items when the initial height of the scrollElemnt is 0. I could fix it by calling virtualizer.measure after updating options.

I created a PR to your branch, so please check it!
https://github.com/s-montigny-desautels/virtual/pull/1

s-montigny-desautels:
@salqueng 
Thanks for your PR. I have adjusted the watch to deeply watch the options object. Since the object doesn't change, the watch function was not called when the scrollElement ref changed.

lucatk:
I have had success using the Vue adapter in one of my projects, and have found no issue.
Is this going to be merged, or is there something missing? @tannerlinsley

salqueng:
@s-montigny-desautels
I just created a PR for your repository! The pr contains fixes related to package information and possible memory leak.

s-montigny-desautels:
Hi! With the help of @salqueng, we've refactored the code to be simpler and safer and added the missing build configurations. Everything should be ready to merge!

s-montigny-desautels:
I've updated to beta 26 and updated package.lock file. CI should pass now :+1:

piecyk:
Thanks @s-montigny-desautels @salqueng 👏

PR #403, opened 2022-10-12
fix: call calculateRange when options change

### LINKED ISSUE #363: Dynamically updating count seems to not trigger an update

piecyk:
fixes #363

suppress  #366

### DIFF
--- a/packages/react-virtual/__tests__/index.test.tsx
+++ b/packages/react-virtual/__tests__/index.test.tsx
@@ -36,7 +36,7 @@ function List({
     observeElementRect: (_, cb) => {
       cb({ height, width })
     },
-    measureElement: () => itemSize ?? 0,
+    measureElement: React.useCallback(() => itemSize ?? 0, [itemSize]),
     rangeExtractor,
   })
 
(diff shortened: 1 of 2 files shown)


piecyk:
Maybe here we should just expose the `calculateRange` and `range` and basic caller will decided when it should recalculate.

Issue #381, opened 2022-08-29
Dynamic height case, when I locate row 5000 and scroll up again, the problem of flickering.


FW-Ezreal:
### Describe the bug

Dynamic height ，There are two problems
1. For example, when I locate row 5000 and scroll up again, the problem of flickering。
2. when I locate row 5000，positioning is not accurate，May be located at 510

### Your minimal, reproducible example

https://codesandbox.io/s/github/tanstack/virtual/tree/beta/examples/react/dynamic?from-embed=&file=/src/main.tsx

### Steps to reproduce

[The official website can also reproduce the problem](https://tanstack.com/virtual/v3/docs/examples/react/dynamic)
1. For example, when I locate row 5000 and scroll up again, the problem of flickering。
2. when I locate row 5000，positioning is not accurate，May be located at 510

### Expected behavior

Help answer the above two questions

### Platform

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36

### tanstack-virtual version

3.0.0-beta.18

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

FW-Ezreal:
<img width="552" alt="image" src="https://user-images.githubusercontent.com/30379963/187116646-2005546b-6570-47af-b983-1c172ef0652a.png">

TheSisb:
I've noticed a similar issue with innacurate smooth scroll on the official examples. This codesandbox works however, and uses a custom build out: https://codesandbox.io/s/fixed-smooth-scroll-for-react-virtual-h0vyjo

filip-kis-fsc:
I'm also experiencing issue no. 1.

It seems to happen specifically when scrolling down a dynamic height list via JS, then scrolling back up via mouse/trackpad (eg. a chat app). If the `estimateSize` function is not completely accurate, scrolling back up will lead to scroll stutter until all items are measured.

This discussion seems related: https://github.com/TanStack/virtual/discussions/195

wuarmin:
@piecyk Do you have an idea about this issue? The [official example](https://tanstack.com/virtual/v3/docs/examples/react/dynamic) shows the issue too.
thanks

piecyk:
[...]

Issue #188, opened 2021-08-16
Changing the size of partially visible items triggers list to scroll 


Karpengold:
According to this line https://github.com/tannerlinsley/react-virtual/blob/433d9663cedb7332d738a236199f9f7e76639039/src/index.js#L178

in case of changing the size of partially visible items (for example expanding) list is scrolling to save same position for partially visible item: https://codesandbox.io/s/react-virtual-resizeobserver-transition-forked-65sg7?file=/src/App.js

https://user-images.githubusercontent.com/17138959/129550951-60a11154-48a6-42cd-a9d6-ccf818ad9bef.mov

piecyk:
Adjusting scroll position when item size change should only happen when we are scrolling backward. Looks like this should fix the this issue https://github.com/tannerlinsley/react-virtual/compare/master...piecyk:fix/adjust-scroll-position-when-item-size-change-only-when-scrolling-backward

Karpengold:
@piecyk cool! As I understood this feature also provide ability to get isScrolling param from react-virtual hook, so I will be able to remove custom scroll handling logic :)

piecyk:
@Karpengold yes, as we want to reset scrollDirection, then having extra isScrolling looked like good idea 🤔 

Let's see what @tannerlinsley will say.

Karpengold:
@piecyk  I found the little hack how to prevent this behaviour 
I use paddingStart property and add some extra space in the beginning (i.e. 1 item height)
And during calculating the start property of each item in the list I decrease it by this extra space

tannerlinsley:
This is a decent workaround for now. We'll work on improving this experience in the next major version for sure.

PR #404, opened 2022-10-12
fix: reduce flickering on scrolling when measuring

piecyk:
@tannerlinsley this should reduce the flickering reported at #381 

Idea here is that rather blocking with scrollCheckFrame, as there can be mulit adjustment till scroll happen accumulate them on pendingScrollOffset.


piecyk:
This is not resolving issues like https://github.com/TanStack/virtual/issues/188 when we don't want to adjust position 🤔

piecyk:
Ok, now i see for what was added scrollCheckFrame, basic preventing adjustment of position when smooth scrolling is happening. 

This has the down side that scroll to index most cases will not land on correct index in case of dynamic sizes.

PR #405, opened 2022-10-13
Observe element for size changes with ResizeObserver

### LINKED ISSUE #376: Virtualizer does not update after adjusting row height

### LINKED ISSUE #391: Dynamic grid maximum update depth exceeded.
### Describe the bug
Dynamic grid does not work combined with react-table.

```
import React from "react";
import {
  flexRender,
  getCoreRowModel,
  useReactTable,
} from "@tanstack/react-table";
import { useVirtualizer } from "@tanstack/react-virtual";


const TableModelGridVirtual = ({ data, columns }) => {
  const table = useReactTable({
    data,
    columns,
    getCoreRowModel: getCoreRowModel(),
  });

  const { rows } = table.getRowModel();

  const parentRef = React.useRef();

    count: rows.length,
    estimateSize: () => 30,
  });

  const columnVirtualizer = useVirtualizer({
    horizontal: true,
    count: columns.length,
    estimateSize: () => 125,
  });

  const halfWay = Math.floor(rows.length / 2);

  return (
    <>
      <div className="py-2 flex gap-2">
        <button onClick={() => rowVirtualizer.scrollToIndex(halfWay)}>
          Scroll to index {halfWay}
        </button>
        <button onClick={() => rowVirtualizer.scrollToIndex(rows.length - 1)}>
          Scroll to index {rows.length - 1}
        </button>
      </div>

      <div
        ref={parentRef}
        className="List"
        style={{
          height: `400px`,
          width: `100%`,
          overflow: "auto",
        }}
      >
        <div
          style={{
            height: rowVirtualizer.getTotalSize(),
            width: columnVirtualizer.getTotalSize(),
            position: "relative",
          }}
        >
          {rowVirtualizer.getVirtualItems().map((virtualRow) => (
            <React.Fragment key={virtualRow.key}>
              {columnVirtualizer.getVirtualItems().map((virtualColumn) => {
                const cell =
                  rows[virtualRow.index].getVisibleCells()[virtualColumn.index];
                console.log(cell);
                return (
                  <div
                    key={virtualColumn.key}
                    ref={(el) => {
                      virtualRow.measureElement(el);
                      virtualColumn.measureElement(el);
                    }}
                    className={
                      virtualColumn.index % 2
                        ? virtualRow.index % 2 === 0
                          ? "ListItemOdd"
                          : "ListItemEven"
                        : virtualRow.index % 2
                        ? "ListItemOdd"
                        : "ListItemEven"
                    }
                    style={{
                      position: "absolute",
                      top: 0,
                      left: 0,
                      transform: `translateX(${virtualColumn.start}px) translateY(${virtualRow.start}px)`,
                    }}
                  >
                    <div
                      style={{
                        height: rows[virtualRow.index],
                        width: columns[virtualColumn.index],
                      }}
                    >
                      {flexRender(
                        cell.column.columnDef.cell,
                        cell.getContext()
                      )}
                    </div>
                  </div>
                );
              })}
            </React.Fragment>
          ))}
        </div>
      </div>
    </>
  );
};

export default TableModelGridVirtual;
```

-

### Steps to reproduce

Does not render

### Expected behavior

Dynamic grid

Always

### Screenshots or Videos

image

### Platform

Windows

### tanstack-virtual version

3.0.0-beta.18

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

piecyk:
Was looking on our options how to add observing element for size changes using ResizeObserver fix #376. What if we would change a bit the api, end expose `measureElement` on top level, not on the item. This would allow use to memo few of them for example in grid and fix #391

Then we nee to bound `measureElement` with index, we can do it via `attribute` by adding `data-index={virtualRow.index}`.

### DIFF
--- a/docs/api/virtual-item.md
+++ b/docs/api/virtual-item.md
@@ -5,13 +5,12 @@ title: VirtualItem
 The `VirtualItem` object represents a single item returned by the virtualizer. It contains information you need to render the item in the cooredinate space within your virtualizer's scrollElement and other helpful properties/functions.
 
 ```tsx
-export interface VirtualItem<TItemElement = unknown> {
+export interface VirtualItem {
   key: string | number
   index: number
   start: number
   end: number
   size: number
-  measureElement: (el: TItemElement | null) => void
 }
 ```
 
@@ -56,11 +55,3 @@ size: number
 ```
 
 The size of the item. This is usually mapped to a css property like `width/height`. Before an item is measured vit the `VirtualItem.measureElement` method, this will be the estimated size returned from your `estimateSize` virtualizer option. After an item is measured (if you choose to measure it at all), this value will be the number returned by your `measureElement` virtualizer option (which by default is configured to measure elements with `getBoundingClientRect()`).
-
-### `measureElement`
-
-```tsx
-measureElement: (el: TItemElement | null) => void
-```
-
-Measures the element using your configured `measureElement` virtualizer option. You are repsonsible for calling this in your virtualizer markup when the component is rendered (eg. using something like React's ref callback prop). By default the `measureElement` virtualizer option is configured to measure elements with `getBoundingClientRect()`.
--- a/docs/api/virtualizer.md
+++ b/docs/api/virtualizer.md
@@ -216,7 +216,7 @@ The current scrollElement for the virtualizer. This property is updated via your
 ### `getVirtualItems`
 
 ```tsx
-type getVirtualItems = () => VirtualItem<TItemElement>[]
+type getVirtualItems = () => VirtualItem[]
 ```
 
 Returns the virtual items for the current state of the virtualizer.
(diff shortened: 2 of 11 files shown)


piecyk:
As note, for example having two lists and keeping only one visible at the time, other hidden by display none. We would like to skip updating sizes for the hidden one as ResizeObserver will call on display change. One solution for this could if virtualizer would expose `measurementsCache` and `indexFromElement`


```tsx
export const useVirtualizer2 = <TScrollElement extends Element, TItemElement extends HTMLElement>({
  updateSize = true,
  ...options
}: VirtualizerOptions<TScrollElement, TItemElement>) => {
    ...options,
    measureElement: (element: TItemElement, instance: Virtualizer<TScrollElement, TItemElement>) => {
      const getSize = () => measureElement(element, instance)

      if (updateSize) {
        return getSize()
      } else {
        const item = instance.measurementsCache[instance.indexFromElement(element)]

        return item ? item.size : getSize()
      }
    },
  })

  return virtualizer
}
```

andylwelch:
Thank you, this is amazing! I can delete so much logic to trigger resize manually

piecyk:
Hi, you can still provide `measureElement` like in my comment above to return item height, basic change here is that the measureElement will be called if ResizeObserver detects change.

piecyk:
@lukesmurray hi, you can still provide `measureElement` like in my comment above to return item height, basic change here is that the measureElement will be called if ResizeObserver detects change.

lukesmurray:
ahh gotcha that makes sense thank you @piecyk.

PR #414, opened 2022-10-25
Trigger recalculation on every update when not scrolling

piecyk:
fixes https://github.com/TanStack/virtual/issues/363

suppress https://github.com/TanStack/virtual/pull/403

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -261,12 +261,15 @@ export interface VirtualizerOptions<
   rangeExtractor?: (range: Range) => number[]
   enableSmoothScroll?: boolean
   scrollMargin?: number
+  scrollingDelay?: number
 }
 
 export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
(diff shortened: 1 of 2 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -391,7 +413,7 @@ export class Virtualizer<TScrollElement = unknown, TItemElement = unknown> {
     },
   )
 
-  private calculateRange = memo(
+  calculateRange = memo(
```
atmgrifter00, line 416: I'm ecstatic that you are trying to solve this, because my team recently just ran head first into this exact issue. My main question with this approach, I think, is why can't we just have `measure` (which is already public) call `calculateRange` as part of its implementation. At the moment, it already doesn't seem to be doing what it semantically says it should be doing (from what I can tell{).

piecyk: Hi, i think we need to re-thing measure, now it's only clears itemMeasurementsCache and trigger update so most cases it will not measure anything. When we merge #405 with adding support observing elements with ResizeObserver i think we could drop it.

PR #440, opened 2022-12-04
[svelte-virtual] Fix  called Virtualizer.setOptions and then Virtualizer to stop working.

### LINKED ISSUE #429: Svelte-adapter infinite list
### Describe the bug

Unable to use setOptions to increase count - when I do this the vrtual listscroll breaks.

https://stackblitz.com/edit/vitejs-vite-j5so1j?file=src%2FApp.svelte

### Steps to reproduce

Scroll down virtual list it will trigger a call to setOptions

### Expected behavior

Virtual list to append new data and enable continuous loading

Every time

### Screenshots or Videos

_No response_

### Platform

windows chrome

### tanstack-virtual version

3.0.0-beta.23

### TypeScript version

_No response_

### Additional context

		"@tanstack/svelte-virtual": "^3.0.0-beta.23",
		"@tanstack/virtual-core": "^3.0.0-beta.23",
		"svelte": "^3.50.1",
		"@sveltejs/kit": "1.0.0-next.544",

### Terms & Code of Conduct

YeungKC:
Currently, the parameter type of `Virtualizer.setOptions` is `Virtualizer<TScrollElement, TItemElement>`, so `onChange` for update virtualizer instance  and the default parameters for `createVirtualizer` and `createWindowVirtualizer` are lost, which is very inconvenient for using svelte-virtual.

Improvement
Now calling `createVirtualizerBase` will provide a setOptions that can set any part of the options, and the other properties are `Virtualizer.options`.

fix #429


YeungKC:
Please review this PR

@hipstermojo @tannerlinsley

piecyk (commented):
@YeungKC what not use options from virtualizer?
```tsx
virtualizer.setOptions({
  ...virtualizer.options,
  count: 2,
})
  ```
  
Could you create minimal codesandbox example?

**Comment by YeungKC** [comment]
> 

If I need set a new onChange and then Virtualizer will be stop.

I think this is a very simple issue. I prefer not to create codesandbox examples. If you are interested, you can check this code.

https://github.com/YeungKC/mvm-site/blob/master/src/routes/(drawer)/transactions/%2Bpage.svelte#L62- L63

**Comment by YeungKC** [comment]
https://stackblitz.com/edit/vitejs-vite-kqnr2s?file=src/App.svelte

I make a codesandbox example.

@piecyk

**Comment by JohnRSim** [comment]
Thanks @YeungKC works great for me.

**Comment by JohnRSim** [comment]
@YeungKC  if I add an image to each row and open chrome tools I'm getting images flicker.
I'm not sure what's causing it atm..  It also happens when browser requires a lot of resource and  chrome tools are closed. 

Svelte
https://stackblitz.com/edit/vitejs-vite-bri3k6?file=src%2FApp.svelte

https://user-images.githubusercontent.com/364208/211013553-7c84d641-35ba-4c07-821d-ef9bc93c72cf.mp4



I'm unable to replicate in react.. :
React
https://codesandbox.io/s/nervous-faraday-j08y0m?file=/src/main.tsx

**Comment by JohnRSim** [comment]
Switched to the official Svelte adapter and upgraded packages .35 and am able to replicate with svelte. 
Any ideas on this?

https://stackblitz.com/edit/vitejs-vite-gwmue4?file=src%2FApp.svelte,package.json

**Comment by JohnRSim** [comment]
ok looks like it is just an issue with 
- Chrome Version 111.0.5529.0 (Official Build) canary (64-bit) 

These 2 works fine on Windows 10 bootcamp
- Firefox 108.02 (64-bit)
- Chrome Version 108.0.5359.125 (Official Build) (64-bit)

I'll do a few more tests 
- but it's weird that I'm unable to replicate the issue at all with React; but can with svelte on canary with chrome tools open..

**Comment by YeungKC** [comment]
I guess this issue is caused by the cache control and the overscan parameter.

When open Chrome tools, default cache control is disable, and the svelte example didn't set the overscan, and then when load a new image of item, it need to reload the image, when the image begin loading, the image size is zero. when it is loaded, the image will have size.

finally, it flash.

**Comment by JohnRSim** [comment]
On Svelte if I added as a test overscan: 50  then any items after 50 flicker for me in svelte? 
I thought overscan does a scan of items before and after - it looks like this might be the issue?
I tried the default of 5 and saw flicker at 25 items 

@piecyk Could you take a quick look and confirm if the issue is with overscan logic where it only scans the initial set of items..
I'm unable to replicate the issue with react works perfectly.

Thanks

**Comment by JohnRSim** [comment]
https://stackblitz.com/edit/vitejs-vite-gwmue4?file=src%2FApp.svelte,package.json

I've updated this example with 20 items first 20 do not flicker - the rest do as I scroll down with the official adapter.
If I change overscan to 50 the full list then none of them flicker I can scroll to the bottom and top without issue.

File: packages/svelte-virtual/src/index.ts [hunk 1]
```diff
@@ -11,29 +11,53 @@ import {
...
+}
 
 function createVirtualizerBase<
   TScrollElement extends Element | Window,
   TItemElement extends Element,
 >(
   options: VirtualizerOptions<TScrollElement, TItemElement>,
-): Readable<Virtualizer<TScrollElement, TItemElement>> {
+): Readable<SvelteVirtualizer<TScrollElement, TItemElement>> {
+  let originalOnChange = options.onChange
```
piecyk, line 31: Why we need originalOnChange here?

packages/svelte-virtual/src/index.ts
```diff
@@ -11,29 +11,53 @@ import {
...
+  setOptions: (
+    options: Partial<VirtualizerOptions<TScrollElement, TItemElement>>,
+  ) => void
+}
 
 function createVirtualizerBase<
   TScrollElement extends Element | Window,
   TItemElement extends Element,
 >(
   options: VirtualizerOptions<TScrollElement, TItemElement>,
```
piecyk, line 29: Let's rename it to initialOptions

packages/svelte-virtual/src/index.ts
```diff
@@ -11,29 +11,53 @@ import {
...
       ...options,
       onChange: (instance: Virtualizer<TScrollElement, TItemElement>) => {
-        set(instance)
-        options.onChange?.(instance)
+        virtualizerWritable.set(instance)
+        originalOnChange?.(instance)
       },
     })
+    virtualizer._willUpdate()
+  }
```
piecyk, line 51: I think we can drop the `originalOnChange` and go with?
```ts
const setOptions = (
  options: Partial<VirtualizerOptions<TScrollElement, TItemElement>>,
) => {
  const resolvedOptions = {
    ...virtualizer.options,
    ...options,
  }
  originalSetOptions({
    ...resolvedOptions,
    onChange: (instance) => {
      virtualizerWritable.set(instance)
      resolvedOptions.onChange(instance)
    },
  })
  virtualizer._willUpdate()
}
  ```

piecyk (commented):
Looks good added few minor comments, thanks @YeungKC 👏

packages/svelte-virtual/src/index.ts
YeungKC: Cause the options type is `Partial<VirtualizerOptions<TScrollElement, TItemElement>>`, and then options maybe didn't have `onChange`, so it need call last `onChange`.

That's why we need `originalOnChange`.

piecyk: I think we don't need it, if merge options with virtualizer.options

PR #466, opened 2023-01-06
fix(virtual-core): round element rect values

piecyk:


salazarm:
whyyyyyyy

salazarm:
the rounding is causing our elements to be too small and their box-shadow gets cut-off by the next element

piecyk:
Hey @salazarm, you can always override observeElementRect with your own implementation if needed!

PR #470, opened 2023-01-10
docs(solid): add example for dynamic usage

Drevoed:
This PR adds a `dynamic` example for usage with SolidJS.

One particularly interesting thing I notices is that for nodes to be measured properly user would have to queue a microtask.

This happens because the ref is initialized ***before*** solid actually sets data attributes. This might of interest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ coverage
 .env.production.local
 .next
 
+.idea
+
 npm-debug.log*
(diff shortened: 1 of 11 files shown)


Drevoed:
Any news on review?

tannerlinsley:
Triaging as out of date. Please reopen a fresh PR if still valid.

PR #472, opened 2023-01-16
Fix the loop call onChange

YeungKC:


YeungKC:
@piecyk Because the PR span is very long, I forgot some details.

YeungKC:
Since core is also beta at present, changes are relatively frequent, and I will not add tests in the near future.

piecyk:
> Since core is also beta at present, changes are relatively frequent, and I will not add tests in the near future.

I think we are pretty close to stable release, also without test we don't know if changes in core broke something in svelte adapter 🤷‍♂️ Was thinking on something simple, like test that the list is rendered.

YeungKC:
> > Since core is also beta at present, changes are relatively frequent, and I will not add tests in the near future.
> 
> 
> 
> I think we are pretty close to stable release, also without test we don't know if changes in core broke something in svelte adapter 🤷‍♂️ Was thinking on something simple, like test that the list is rendered.

Actually the svelte, react, vue version is very simple, if the core code already has test code to ensure stability, most of the time is stable, if the core code is already a stable version, I will consider add test code to the svelte code.

PR #486, opened 2023-01-26
Fix stale range when updating options

### LINKED ISSUE #485: Virtual list is re-rendered with a stale range initially when changing options
### Describe the bug

When changing `options.count` the list is re-rendered once with a stale range and is then re-rendered again with an updated range. It can lead to a flash of content.

Preview: https://81gdxs.sse.codesandbox.io/
Edit: https://codesandbox.io/s/brave-firefly-81gdxs

### Steps to reproduce

1. Go to https://81gdxs.sse.codesandbox.io/
2. Open the browser debugger
3. Click on the button "20" in the page
4. The debugger will pause the execution, at this point you can observe that the list has been re-rendered with the new rows. But it doesn't render enough items to fill the visible area.
5. Resume execution, the list will be re-rendered with enough items to fill the visible area.

### Expected behavior

What is strange is that it does render with the new elements but some numbers are out of date. So the list is partially updated.

I would expect the list to re-render with everything updated in a single render. Or at least not render with partially updated data.

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/2291025/214835443-5d387247-db02-45bf-9e40-1e66de4c31d0.mov

### Platform

- OS: MacOS
- Browser: Brave

### tanstack-virtual version

3.0.0-beta.39

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

zhouzi:
As seen in #485 when the list re-renders, it resolves the new indexes based on a stale range. The goal of this PR is to make sure indexes are always resolved based on an updated range.

Not sure if it defeats the optimizations made in #353. From what I understand most methods are memoized but `.notify()` may end up being called too often?

Fixes #485

### DIFF
--- a/packages/react-virtual/src/index.tsx
+++ b/packages/react-virtual/src/index.tsx
@@ -38,7 +38,6 @@ function useVirtualizerBase<
   )
 
   instance.setOptions(resolvedOptions)
-  instance.calculateRange()
 
   React.useEffect(() => {
     return instance._didMount()
--- a/packages/react-virtual/src/index.tsx
+++ b/packages/react-virtual/src/index.tsx
@@ -38,6 +38,7 @@ function useVirtualizerBase<
   )
 
   instance.setOptions(resolvedOptions)
+  instance.calculateRange()
 
   React.useEffect(() => {
     return instance._didMount()
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -514,7 +514,7 @@ export class Virtualizer<
   private getIndexes = memo(
     () => [
       this.options.rangeExtractor,
-      this.range,
+      this.calculateRange(),
       this.options.overscan,
       this.options.count,
     ],


packages/react-virtual/src/index.tsx
```diff
@@ -38,6 +38,7 @@ function useVirtualizerBase<
   )
 
   instance.setOptions(resolvedOptions)
+  instance.calculateRange()
```
zhouzi, line 41: With this change the list renders the appropriate items when changing `options.count` in a single render:

https://user-images.githubusercontent.com/2291025/214837437-5a7815fd-da62-4658-8343-7e47d1057783.mov

But I am not sure if that's the best way to fix it for two reasons:

1. Calling `instance.setOptions` and/or `instance.calculateRange` in the body of the component seems risky in regards to React's concurrent mode. React can start a render cycle, which in this case would update the virtual instance's options and data, and abort that cycle. If that happens, the virtual instance would have been updated despite the abort, leading to mismatches between React's state and virtual's. I could be wrong though but that's how I understand it. The fix here would be to wrap them in a `useEffect`/`useLayoutEffect` (it'd be working slightly differently though by delaying updates).
2. I wonder if calling `calculateRange` every time the component renders wouldn't defeat other optimizations. From what I understand most things are memoized anyway but perhaps that wouldn't be ideal.

PR #487, opened 2023-01-27
fix: calculate range when getting items

### LINKED ISSUE #485: Virtual list is re-rendered with a stale range initially when changing options

piecyk:
closes #485

### DIFF
--- a/packages/react-virtual/__tests__/index.test.tsx
+++ b/packages/react-virtual/__tests__/index.test.tsx
@@ -42,6 +42,8 @@ function List({
 
   const measureElement = dynamic ? rowVirtualizer.measureElement : undefined
 
+  const items = rowVirtualizer.getVirtualItems()
+
   return (
     <div
       ref={parentRef}
@@ -55,7 +57,7 @@ function List({
           position: 'relative',
         }}
       >
-        {rowVirtualizer.getVirtualItems().map((virtualRow) => (
+        {items.map((virtualRow) => (
           <div
             data-testid={`item-${virtualRow.key}`}
             key={virtualRow.key}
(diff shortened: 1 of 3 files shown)


packages/react-virtual/__tests__/index.test.tsx
```diff
@@ -128,12 +130,12 @@ test('should render given dynamic size after scroll', () => {
...
-  expect(screen.queryByText('Row 4')).toBeInTheDocument()
-  expect(screen.queryByText('Row 9')).toBeInTheDocument()
-  expect(screen.queryByText('Row 10')).not.toBeInTheDocument()
+  expect(screen.queryByText('Row 2')).not.toBeInTheDocument()
+  expect(screen.queryByText('Row 3')).toBeInTheDocument()
+  expect(screen.queryByText('Row 6')).toBeInTheDocument()
+  expect(screen.queryByText('Row 7')).not.toBeInTheDocument()
 
-  expect(renderer).toHaveBeenCalledTimes(2)
+  expect(renderer).toHaveBeenCalledTimes(3)
```
zhouzi, line 138: I am not familiar with these tests but curious why it's rendering one more time than before? I'd think it would render less.

zhouzi:
Nice, looks like an elegant solution 👍

packages/react-virtual/__tests__/index.test.tsx
piecyk: As in previous test range was calculated only one time when scroll jumped, now we do it again as in dynamic case changing sizes should trigger checking the range.

maarekj:
What version is this fix in?
Thank you

PR #497, opened 2023-02-07
refactor: observer methods

piecyk:


samouri:
Hey @piecyk,

I was reading through `virtual-core/src/index.js` yesterday and was confused by this change.
Any chance you can help me understand why you switched from using the data within the `ResizeObserver` callback to always calling `getBoundingClientRect`? 

My impression was that `getBcr` is initially necessary because of how ResOb is async. After that it is significantly more performant to rely on `ResizeObserver`. Let me know if I'm missing something!

Thanks,
Jake

samouri:
@piecyk: the code is definitely cleaner. I'm not actually a consumer of `react-virtual`, but I'm concerned that [layout thrashing](https://web.dev/avoid-large-complex-layouts-and-layout-thrashing/#avoid-forced-synchronous-layouts) is going to start appearing in folk's profiles.

The ResizeObserver API is sort-of a shame in that 99% of the time folks want an API like `resizeObserver.observe(element, callback)` instead of the actual API. That would make it much easier to do the right thing here instead of synchronously recomputing the values/forcing relayout.

piecyk:
> but I'm concerned that [layout thrashing](https://web.dev/avoid-large-complex-layouts-and-layout-thrashing/#avoid-forced-synchronous-layouts) is going to start appearing in folk's profiles.

that's great @samouri that's for checking this 🙇  Regarding layout thrashing, i think it's not that simple. Did some testing when implementing dynamic size list, observing each item with ResizeObserver and then update list when dimensions changes https://github.com/TanStack/virtual/blob/beta/packages/virtual-core/src/index.ts#L267-L283

Overall the numbers was the same in google chrome dev tools using the `content rect` vs calling getBoundingClientRect, that's why went with latter 🤔

piecyk:
@samouri as conclusion from this discussion created a PR https://github.com/TanStack/virtual/pull/500 looking forward to how it really works, as from quick testing it's not noticeable.

PR #500, opened 2023-02-08
refactor: use ResizeObserver entry

piecyk:
TODO: compare perfomance using content rect ResizeObserver callback to calling getBoundingClientRect, will the [layout thrashing] really happen (https://web.dev/avoid-large-complex-layouts-and-layout-thrashing/#avoid-forced-synchronous-layouts)

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -67,17 +67,26 @@ export const observeElementRect = <T extends Element>(
     return
   }
 
-  const handler = () => {
-    const { width, height } = element.getBoundingClientRect()
+  const handler = (rect: { width: number; height: number }) => {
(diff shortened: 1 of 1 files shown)


piecyk:
Looks like calling getBoundingClientRect in ResizeObserver callback will not trigger `layout thrashing` at least not in chrome/firefox, it's same as reading from callback entry.

Hard to get exact numbers as they are pretty close, created simple tests case that increased width by 5 every 100ms from 200 to 800, to keep it simple, items have fixed width. 

for example, on left we are having latest beta, on right reading values from entry  
Zrzut ekranu 2023-02-9 o 20 57 38

I think will merge this PR, overall by passing entry to measureElement user can now know that they are in RO callback, 

cc @samouri

PR #501, opened 2023-02-09
fix (vue-virtual): Update peer dependencies to take into account any …

PerpetualWar:
…2.7 or greater minor version or 3.0 or greater minor version


piecyk:
Thanks 👍

PR #516, opened 2023-02-26
fix: replace typeof window checks with typeof document

redabacha:
currently the `@tanstack/react-virtual` package throws `useLayoutEffect does nothing on the server` warnings during ssr when using a deno runtime. this is because deno has a global window object available for browser compatibility (though is being considered for removal https://github.com/denoland/deno/issues/13367). also see [here](https://remix.run/docs/en/v1/pages/gotchas#typeof-window-checks) for further reading.

this pr replaces all `typeof window` checks across all packages with `typeof document` instead which works in all environments.

prior art: https://github.com/framer/motion/pull/1522

### DIFF
--- a/packages/vue-virtual/src/index.ts
+++ b/packages/vue-virtual/src/index.ts
@@ -103,7 +103,8 @@ export function useWindowVirtualizer<TItemElement extends Element>(
   return useVirtualizerBase<Window, TItemElement>(
     computed(() => ({
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      getScrollElement: () => (typeof Window !== 'undefined' ? window : null!),
+      getScrollElement: () =>
+        typeof document !== 'undefined' ? window : null!,
       observeElementRect: observeWindowRect,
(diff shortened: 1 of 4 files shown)


packages/vue-virtual/src/index.ts
```diff
@@ -103,7 +103,8 @@ export function useWindowVirtualizer<TItemElement extends Element>(
   return useVirtualizerBase<Window, TItemElement>(
     computed(() => ({
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      getScrollElement: () => (typeof Window !== 'undefined' ? window : null!),
```
redabacha, line 106: this would have always evaluated to `null` since it should have been lowercase `window` instead of `Window`

redabacha, line 106: this would have always evaluated to `null` since it should have been lowercase `window` instead of `Window` :sweat_smile:

piecyk (approved):

piecyk:
Thanks @redabacha

PR #518, opened 2023-02-26
Add unsupported type guard for Safari `borderBoxSize`

zi-gae:
## Descriptions
Currently, the borderBoxSize property in ResizeObserverEntry is not supported on Safari. This can cause errors and unexpected behavior in applications that rely on this property. This pull request proposes the addition of an unsupported type guard for Safari's borderBoxSize to prevent errors and improve the user experience.

## Changes
Added optional chaining to the borderBoxSize type guard in ResizeObserverEntry for browsers that do not support this property.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -76,7 +76,7 @@ export const observeElementRect = <T extends Element>(
 
   const observer = new ResizeObserver((entries) => {
     const entry = entries[0]
-    if (entry) {
+    if (entry?.borderBoxSize?.[0]) {
       const box = entry.borderBoxSize[0]
       if (box) {
         handler({ width: box.inlineSize, height: box.blockSize })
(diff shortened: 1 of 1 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -76,7 +76,7 @@ export const observeElementRect = <T extends Element>(
 
   const observer = new ResizeObserver((entries) => {
     const entry = entries[0]
-    if (entry) {
+    if (entry?.borderBoxSize?.[0]) {
```
piecyk, line 79: Maybe checking if `borderBoxSize` exist in entry would be enough?
```suggestion
    if (entry?.borderBoxSize) {
```

piecyk: @zi-gae can you update this, checking if it exist will be way faster.

zi-gae: thank you for your suggestion. https://github.com/TanStack/virtual/pull/518/commits/242ca34fdff518b7bb553620286548d2571e93a7

PR #521, opened 2023-02-27
feat: add elementKeyAttribute, controls element cache

piecyk:


chrisdhanaraj:
Relatively sure this would solve my issue over here as well - https://github.com/TanStack/virtual/issues/533. I could declare a different element key and have it recompute the values

Issue #30, opened 2020-05-22
Dynamic masonry component


jediswear:
Is there a way to build like masonry component? Each col height should depend on children height.
Something like https://bvaughn.github.io/react-virtualized/#/components/Masonry

o-alexandrov:
There is an example in #6 
@lishine proposed two choices:
- one with positioning adjustments he already showcases in his [github repo](https://github.com/lishine/unsplash-search/blob/dev/src/Gallery/view/List/InfiniteListWindow.tsx) for react-window
- another approach is to add multiple virtualized columns (multiple lists in other words)
  - this way you add an event listener on a wrapper that all of them will be scrolled on `mousewheel` event

It'd be great to know @tannerlinsley 's view :)

lishine:
> There is an example in #6
> @lishine proposed two choices:
> 
> * one with positioning adjustments he already showcases in his [github repo](https://github.com/lishine/unsplash-search/blob/dev/src/Gallery/view/List/InfiniteListWindow.tsx) for react-window
> * another approach is to add multiple virtualized columns (multiple lists in other words)
>   
>   * this way you add an event listener on a wrapper that all of them will be scrolled on `mousewheel` event
> 
> It'd be great to know @tannerlinsley 's view :)

Exactly.
And this lib variant:
https://github.com/lishine/unsplash-search/blob/dev/src/Gallery/view/List/InfiniteListVirtual.tsx

tannerlinsley:
This issue has been automatically closed due to inactivity. If this issue represented an unfulfilled feature request, please start a new discussion instead of an issue. If this issue needs more attention despite its age, please comment to reopen, or open a new issue with updated information. Thanks!

zhouzi:
Just a heads up for the ones looking for a masonry layout, it has been added to the library. It can be seen in action in the variable examples: https://tanstack.com/virtual/v3/docs/examples/react/variable

misaelvillaverde:
Hello, the example doesn't address for dynamic height, using the virtualizer.measureElement function in the ref doesn't work when using lanes, the items use the estimated size and overlaps each other, is there an up to date example?

PR #523, opened 2023-03-01
feat: add support for masonry

zhouzi:
This PR aims at adding support to masonry layouts as expressed in #30

In terms of logic the difference is that instead of taking the previous element in the list, it looks for the one that's the furthest away from the end of the container.

For example in a top to bottom list with two columns and these elements:

```
----------- -----------
|    1    | |    2    |
|         | -----------
-----------
```

The third element should be inserted below box 2 and not box 1 because box 2 is the one that's the furthest from the end of the container.

```
----------- -----------
|    1    | |    2    |
|         | -----------
----------- -----------
            |    3    |
            |         |
            -----------
```

The fourth element would be inserted below box 1.

I updated the docs and added an example to the variable examples:

https://user-images.githubusercontent.com/2291025/222973210-ecba891e-3848-4093-8b2c-79c1b901b017.mov

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -34,7 +34,7 @@ export interface VirtualItem {
   start: number
   end: number
   size: number
-  column: number
+  lane: number
 }
 
 interface Rect {
@@ -262,7 +262,7 @@ export interface VirtualizerOptions<
   scrollingDelay?: number
   indexAttribute?: string
   initialMeasurementsCache?: VirtualItem[]
-  columns?: number
+  lanes?: number
 }
 
 export class Virtualizer<
@@ -347,7 +347,7 @@ export class Virtualizer<
       scrollingDelay: 150,
       indexAttribute: 'data-index',
       initialMeasurementsCache: [],
-      columns: 1,
+      lanes: 1,
       ...opts,
     }
   }
@@ -473,29 +473,29 @@ export class Virtualizer<
         for (let m = i - 1; m >= 0; m--) {
           const measurement = measurements[m]!
 
-          if (furthestMeasurementsFound.has(measurement.column)) {
+          if (furthestMeasurementsFound.has(measurement.lane)) {
             continue
           }
 
           const previousFurthestMeasurement = furthestMeasurements.get(
-            measurement.column,
+            measurement.lane,
           )
           if (
             previousFurthestMeasurement == null ||
             measurement.end > previousFurthestMeasurement.end
           ) {
-            furthestMeasurements.set(measurement.column, measurement)
+            furthestMeasurements.set(measurement.lane, measurement)
           } else if (measurement.end < previousFurthestMeasurement.end) {
-            furthestMeasurementsFound.set(measurement.column, true)
+            furthestMeasurementsFound.set(measurement.lane, true)
           }
 
-          if (furthestMeasurementsFound.size === this.options.columns) {
+          if (furthestMeasurementsFound.size === this.options.lanes) {
             break
           }
         }
 
         const furthestMeasurement =
-          furthestMeasurements.size === this.options.columns
+          furthestMeasurements.size === this.options.lanes
             ? Array.from(furthestMeasurements.values()).sort(
                 (a, b) => a.end - b.end,
               )[0]
@@ -513,9 +513,9 @@ export class Virtualizer<
 
         const end = start + size
 
-        const column = furthestMeasurement
-          ? furthestMeasurement.column
-          : i % this.options.columns
+        const lane = furthestMeasurement
+          ? furthestMeasurement.lane
+          : i % this.options.lanes
 
         measurements[i] = {
           index: i,
@@ -523,7 +523,7 @@ export class Virtualizer<
           size,
           end,
           key,
-          column,
+          lane,
         }
       }
 
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -458,10 +458,6 @@ export class Virtualizer<
     measurements: VirtualItem[],
     index: number,
   ) => {
-    if (this.options.lanes === 1) {
-      return measurements[index - 1]
-    }
-
     const furthestMeasurementsFound = new Map<number, true>()
     const furthestMeasurements = new Map<number, VirtualItem>()
     for (let m = index - 1; m >= 0; m--) {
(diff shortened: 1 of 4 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -34,6 +34,8 @@ export interface VirtualItem {
   start: number
   end: number
   size: number
+  top: number
+  left: number
```
zhouzi, line 38: Until now the naming for the properties was orientation-agnostic:

- `start` would either refer to the element's x or y coordinate depending on `options.horizontal`
- `size` would either refer to the element's width or height
- `end` is `start` + `size` so it would either refer to x2 or y2

I can see how it makes the code simpler in the library in some places but I feel like it can be confusing for the end user (e.g `translateY(${row.start})`).

Furthermore, with this PR elements are positioned on both axis and I don't know what name would make sense for the opposite axis. I'll think about it but maybe it's fine to have `start` as well as `top` and `left`.

packages/virtual-core/src/index.ts
```diff
@@ -261,6 +263,8 @@ export interface VirtualizerOptions<
   scrollingDelay?: number
   indexAttribute?: string
   initialMeasurementsCache?: VirtualItem[]
+  columns?: number
+  columnsSize?: number
```
zhouzi, line 267: Not a big fan of having these as separate properties since they depend on each other and it means users you pass `columns` and not `columnsSize` or `columnsSize` and not `columns`.

An alternative would be `columns: number[]` where `number` would be the column's size. It would also make it possible to have columns with different sizes.

With that being said, having the column's size as a `number` closes the gate to variable columns size. For example you could have three columns, the one's `30%`, the second `30%` and the third `40%`. So another solution could be to have a `estimateColumnSize: (index: number) => number`.

piecyk:
Thanks @zhouzi for this, will check it out 👍 Wondering how big impact on getMeasurements performance would this calculations do.

> Box 3 should go below box 1 because it's the one that's the furthest from the end of the container (assuming a vertical list from top to bottom).

Ooo, didn't expect that 😅

packages/virtual-core/src/index.ts
```diff
@@ -464,16 +470,68 @@ export class Virtualizer<
...
+            continue
+          }
+
+          const previous = columns.get(index)
+          if (previous == null || measurement.end > previous.end) {
+            columns.set(index, measurement)
+          } else if (measurement.end < previous.end) {
+            complete.set(index, true)
+          }
+        }
```
zhouzi, line 496: > Wondering how big impact on getMeasurements performance would this calculations do.

@piecyk I did a few things to try to minimize the performance impact on the calculations. The first thing is that it loops over the elements starting from the end since the element we are looking for is very likely to be among the lasts. The second thing is that once it found a column's last element, it ignores other elements from that column. Finally, it stops looping over the elements when every columns' last element has been found.

For lists not using masonry (so only 1 column), it will always loop only over the last element so it should be almost as fast as before (I think).

With that being said I haven't run benchmarks and I may be missing the implication of this change on other parts of the code.

packages/virtual-core/src/index.ts
zhouzi: Turns out there's no need for top/left so I removed them.

packages/virtual-core/src/index.ts
zhouzi: In the end the columns' size is entirely handled in user-land so we just need `columns: number`.

zhouzi:
@piecyk this PR is now ready for review 🙏 I updated the docs, added an example and updated the PR's description (and updated recording).

tannerlinsley:
Can masonry be used horizontally? And if so, how  about renaming column to “lane” So it’s orientation agnostic?

zhouzi:
> Can masonry be used horizontally?

Yes it can, although I think they're pretty uncommon. But since the library abstracts the orientation nicely, it just works. I added an example:

https://user-images.githubusercontent.com/2291025/222977502-1e73cc91-eb06-40c2-8b62-f1a93d83a761.mov

*Note: in the recording we can see items disappearing when scrolling in the vertical masonry. That's because the overscan is low and with masonry there are more items shown at the same time.*

> how about renaming column to “lane” So it’s orientation agnostic?

I like this and went with it: I changed column to lane.

packages/virtual-core/src/index.ts
```diff
@@ -464,16 +467,64 @@ export class Virtualizer<
 
       for (let i = min; i < count; i++) {
         const key = getItemKey(i)
-        const measuredSize = itemSizeCache.get(key)
-        const start = measurements[i - 1]
-          ? measurements[i - 1]!.end
+
```
piecyk, line 470: I would do this calculations in condition

```tsx
const measuredSize = itemSizeCache.get(key)
const size =
  typeof measuredSize === 'number'
    ? measuredSize
    : this.options.estimateSize(i)

if (this.options.columns > 1) {
  // ... calculations

  const start = furthestMeasurement
    ? furthestMeasurement.end
    : paddingStart + scrollMargin

  const end = start + size

  const column = furthestMeasurement
    ? furthestMeasurement.column
    : i % this.options.columns

  measurements[i] = {
    index: i,
    start,
    size,
    end,
    key,
    column,
  }
} else {
  const start = measurements[i - 1]
    ? measurements[i - 1]!.end
    : paddingStart + scrollMargin
  const end = start + size

  measurements[i] = {
    index: i,
    start,
    size,
    end,
    key,
    column: 1,
  }
}
```

piecyk: Or even better lift the check `this.options.lanes > 1` before the for loop, i think here is pretty acceptable to have this small duplication. Overall @zhouzi awesome work 👏

zhouzi: The only difference between the two code branches is how the previous measurement is found. So I added a method that follows your suggestion by returning `measurements[i - 1]` when lanes === 1. No code duplication 👍

piecyk: Ok, that work also, thanks 👍 Imho having check for `lanes` before calling the method would better, as there is no need to call it, also logic is one place.

Will test it a bit and lets merge it 💪

zhouzi: > Imho having check for `lanes` before calling the method would better, as there is no need to call it, also logic is one place.

Done, I moved the condition out of the method 😉

jawadsef:
Hey @zhouzi @piecyk, would it be possible to add an example with dynamic height for individual items inside the vertical masonry? I currently have major issues when I try to enable dynamic height for the vertical masonry, the smooth scrolling doesn't work and sometimes it can take a few seconds before any items are visible after scrolling.

zhouzi:
Hey @jawadsef, there's an example of a vertical masonry with variable item height here: https://tanstack.com/virtual/v3/docs/examples/react/variable

If you can create a sandbox reproducing the issues you are facing, I'd be happy to look into it.

PR #558, opened 2023-06-11
Add force option in observeElementOffset callback for more intricate scrolling requirements

nkgm:
In practice, this is a tiny addition to the memoization mechanism which could come in handy for more use-cases in the future.

The main drive behind it was to enable a smooth push-out transition between sections, demonstrated in this [CodeSandbox](https://codesandbox.io/p/sandbox/crazy-shadow-lllc2y) based on the original Sticky example. The existing mechanism optimizes updates based on `range` changes, but in this case, the two section headers collide "mid-item" making the transition inconsistent and jumpy (even more so when having different item sizes per type of row, like in this example).

Another option would be to make `Virtualizer.notify` public, but this solution plays nicer with `observeElementOffset` and feels more principled.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -179,7 +179,7 @@ export const measureElement = <TItemElement extends Element>(
   }
   return Math.round(
     element.getBoundingClientRect()[
-      instance.options.horizontal ? 'width' : 'height'
+    instance.options.horizontal ? 'width' : 'height'
     ],
   )
 }
@@ -237,7 +237,7 @@ export interface VirtualizerOptions<
   ) => void | (() => void)
   observeElementOffset: (
     instance: Virtualizer<TScrollElement, TItemElement>,
-    cb: (offset: number) => void,
+    cb: (offset: number, force?: boolean) => void,
   ) => void | (() => void)
 
   // Optional
@@ -340,7 +340,7 @@ export class Virtualizer<
       horizontal: false,
       getItemKey: defaultKeyExtractor,
       rangeExtractor: defaultRangeExtractor,
-      onChange: () => {},
+      onChange: () => { },
       measureElement,
       initialRect: { width: 0, height: 0 },
       scrollMargin: 0,
@@ -398,7 +398,7 @@ export class Virtualizer<
       )
 
       this.unsubs.push(
-        this.options.observeElementOffset(this, (offset) => {
+        this.options.observeElementOffset(this, (offset, force) => {
           this.scrollAdjustments = 0
 
           if (this.scrollOffset === offset) {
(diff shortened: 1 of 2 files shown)


piecyk:
Thanks @nkgm, imho also one options would be, as rangeExtractor is part of memo include some key in deps that will force calling rangeExtractor.

Had a quick look on your sticky example, kinda complicated 😅 maybe we can simplify it a bit 🤔

nkgm:
Hey @piecyk, as a Scala dev myself 🤘, I'd be even more excited to get your feedback and simplify the example further (I can see a few candidates already, but I'd love to get your take on what should/could be made more simple). If it all works out, we could add it to the list of examples.

On this PR, can you share any concerns, and discuss a bit further the proposed *key in deps* alternative? IMO the current approach is a safe/sensible addition that could even stand on its own as an internal api improvement. It also works quite nice for its intended use-case, keeping everything within the confines of the `Virtualizer` avoiding additional scroll listeners. Then again, your knowledge of the library is way more thorough, so I'm looking forward to your feedback!

piecyk:
@nkgm will this be the same? https://codesandbox.io/p/sandbox/thirsty-cloud-kds8wv?file=%2Fsrc%2Fmain.tsx%3A64%2C45

nkgm:
Ok first things first, this is nothing short of genius, it's way simpler, and no doubt performs as optimal as can be. I also considered the height hack early on, with the only difference of having all sticky indexes nested in `<div>`s like you did for the active one, and using `position: sticky; transform: none` for the actual headers, but decided to go for *flexibility* first. Let's say for example that you wanted to add a conditional style or animation to the headers at the precise moment they come in contact with each other, *or* that for some reason you cannot afford to nest the headers and tweak the height, or some other crazy requirement your users are sure to come up with? This PR's ambition is not just solving this particular problem, but *any* problem that requires updating the component at any random scroll position.

piecyk:
> This PR's ambition is not just solving this particular problem, but any problem that requires updating the component at any random scroll position.

Yeah, i agree. Just don't sure if we need to add `force`, as you can just re-render by calling the onChange, in your example would be something like.

```tsx
cb(scrollOffset, false);
if (force) {
  console.log(">>> Forced Update", scrollOffset);
  instance.options.onChange?.(instance);
}
```

nkgm:
Yah pretty much same thing I guess. Please consider refactoring into a dedicated method on the Virtualizer instance without the optional chaining. I'd still like `force` in the `observeElementOffset` callback for convenience, but yours is the more general case and it gets the job done, so feel free to close.

Thanks for your help!

tannerlinsley:
This PR is stale. Happy to continue discussing in a new PR

PR #598, opened 2023-10-03
fix: prevent returning items when container size is zero

### LINKED ISSUE #596: Dynamic item size uses estimateSize fallback when inside portal with dynamically created element
### Describe the bug

I don't know if this is a bug that can be fixed on your side but I experienced a strange behavior when rendering my virtualized list inside a portal. This occurs when I create an anchor for the portal inside `useEffect` and use the node created with `document.createElement`. If I use `document.body` or something reachable with `document.getElementById` instead, it works.

Context: I am using a ComboBox and render the list inside a dropdown or inside a modal on mobile devices. I did not notice it at first because in my development workflow the items usually get fetched after initially rendering the list. These items get cached based on the query so they don't have to fetched again. You can simulate this behavior by changing the default value in the example.

PS: Thanks for creating and maintaining this library!

https://codesandbox.io/p/sandbox/vigilant-kirch-v8q98r

### Steps to reproduce

1. Press "Open modal"
=> First items are squished
2. Change `createPortal` container to `document.body`
3. Reload and press "Open modal"
=> First items are not squished

Lazy: Change [default state in the options](https://codesandbox.io/p/sandbox/vigilant-kirch-v8q98r?embed=1&file=%2Fsrc%2Fcomponents%2FList%2FList.tsx%3A18%2C1-19%2C1) from `_options` to `null`
1. Press "Open modal"
=> First items are not squished

### Expected behavior

Items are not squished

Every time

### Screenshots or Videos

image

### Platform

OS: Windows 11
Browser: Google Chrome 117.0.5938.132 64-Bit
React: 17.0.2 and 18.2.0

### tanstack-virtual version

3.0.0-beta.61

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

piecyk:
closes #596

### DIFF
--- a/examples/react/dynamic/src/main.tsx
+++ b/examples/react/dynamic/src/main.tsx
@@ -74,7 +74,7 @@ function RowVirtualizerDynamic() {
               top: 0,
               left: 0,
               width: '100%',
-              transform: `translateY(${items[0].start}px)`,
+              transform: `translateY(${items[0]?.start ?? 0}px)`,
             }}
           >
             {items.map((virtualRow) => (
@@ -131,7 +131,7 @@ const RowVirtualizerDynamicWindow = () => {
             left: 0,
             width: '100%',
             transform: `translateY(${
-              items[0].start - virtualizer.options.scrollMargin
+              items[0] ? items[0].start - virtualizer.options.scrollMargin : 0
             }px)`,
           }}
         >
--- a/packages/react-virtual/__tests__/index.test.tsx
+++ b/packages/react-virtual/__tests__/index.test.tsx
@@ -148,7 +148,7 @@ test('should render given dynamic size after scroll', () => {
   expect(screen.queryByText('Row 6')).toBeInTheDocument()
   expect(screen.queryByText('Row 7')).not.toBeInTheDocument()
 
-  expect(renderer).toHaveBeenCalledTimes(3)
+  expect(renderer).toHaveBeenCalledTimes(2)
 })
 
 test('should use rangeExtractor', () => {
--- a/packages/virtual-core/__tests__/index.test.ts
+++ b/packages/virtual-core/__tests__/index.test.ts
@@ -3,3 +3,15 @@ import { Virtualizer } from '../src/index'
 test('should export the Virtualizer class', () => {
   expect(Virtualizer).toBeDefined()
 })
+
+test('shoul', () => {
+  const virtualizer = new Virtualizer({
+    count: 100,
+    getScrollElement: () => null,
+    estimateSize: () => 50,
+    scrollToFn: jest.fn(),
+    observeElementRect: jest.fn(),
+    observeElementOffset: jest.fn(),
+  })
+  expect(virtualizer.getVirtualItems()).toEqual([])
+})
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -587,13 +587,16 @@ export class Virtualizer<
       this.calculateRange(),
       this.options.overscan,
       this.options.count,
+      this.getSize(),
     ],
-    (rangeExtractor, range, overscan, count) => {
-      return rangeExtractor({
-        ...range,
-        overscan,
-        count,
-      })
+    (rangeExtractor, range, overscan, count, outerSize) => {
+      return outerSize === 0
+        ? []
+        : rangeExtractor({
+            ...range,
+            overscan,
+            count,
+          })
     },
     {
       key: process.env.NODE_ENV !== 'production' && 'getIndexes',


wuarmin:
Hey @piecyk ,
since this change, the following code no longer works for me.

```javascript
  useEffect(() => {
    if (data.length === 0) {
      return;
    }

    const firstSelectedRow = Object.keys(cache.getSelectedRows())[0];
    if (firstSelectedRow) {
      rowVirtualizer.scrollToIndex(firstSelectedRow, {
        align: "center",
        smoothScroll: false,
      });
    }
  }, [data, cache, rowVirtualizer]);
```
The code is in my table component and ensures that the last selected row is visible for the user. This worked fine with `v3.0.0-beta.61`, and no longer works with version with `v3.0.0-beta.61`.

Can you please help me?

thanks

piecyk:
@wuarmin for keeping selected rows pease use range extractor 

```tsx
import { defaultRangeExtractor, useVirtualizer } from '@tanstack/react-virtual'

const Foo = () => {
  const selectedIndexes = [0, 50]

    count: 100,
    getScrollElement: () => ref.current,
    estimateSize: () => 50,
    rangeExtractor: React.useCallback((range: Range) => {
      if (selectedIndexes.length === 0) {
        return defaultRangeExtractor(range)
      } else {
        const next = new Set([...selectedIndexes, ...defaultRangeExtractor(range)])
        return [...next].sort((a, b) => a - b)
      }
    }, [selectedIndexes]),
  })
  
  return ...
}
```

piecyk:
@wuarmin ooo that is not good, could you add an issue and create codesandbox example? This need to be fixed before v3 release.

piecyk:
@wuarmin I think i see the issue, for now you can wrap the scrollToIndex with requestAnimationFrame

```tsx
requestAnimationFrame(() => {
  virtualizer.scrollToIndex(100)
})
```

wuarmin:
@piecyk Thank you very much! That works. Please, could you explain, why `requestAnimationFrame` is necessary?

Thanks

piecyk:
It's for couple of reasons, mainly the scroll to index when we have dynamic case is not trivial. 

When jumping to index after items are mounted we need to read actual size compare it with estimated and update the list if needed, then the position of the index will change, long story short on lib side we do it in loop ( pushing to next tick, so the browser will have time to render items and we can read the size ) it's enabled when we detect dynamic mode. 

Basic changes showed that the how we detect dynamic mode is not to robust ( we check if we have the elements in cache )

> Could you please briefly explain, why requestAnimationFrame is necessary?

to push the scrolling to next tick, when we have the elements rendered.

FBNitro:
I have this working on my application, I just had to add the ?? 0 to the `rowVirtualizer.getVirtualItems()[0]?.start ?? 0` line, similar to what was done on the PR update to the samples.

But this change made testing with react-testing-library much more difficult because it doesn't actually render in a browser.  The actual size of the containers is 0.  If you have any suggestions @piecyk please let me know.  Options like using cypress are not available to me.  I'll try a few things, but open to suggestions.  It would be like similar to writing a unit test for the https://tanstack.com/virtual/v3/docs/examples/react/dynamic list example.

piecyk:
> I have this working on my application, I just had to add the ?? 0 to the `rowVirtualizer.getVirtualItems()[0]?.start ?? 0` line, similar to what was done on the PR update to the samples.
> 
> But this change made testing with react-testing-library much more difficult because it doesn't actually render in a browser. The actual size of the containers is 0. If you have any suggestions @piecyk please let me know. Options like using cypress are not available to me. I'll try a few things, but open to suggestions. It would be like similar to writing a unit test for the https://tanstack.com/virtual/v3/docs/examples/react/dynamic list example.

@FBNitro that is true, for unit test you can mock the getBoundingClientRect for the scroller element, or provide observeElementRect when testing.

FBNitro:
Ah great, thank you for guiding me in the correct direction.  I updated the getBoundingClientRect in the htmlElement prototype and that worked for me.

```
const DEFAULT_BOUNDING_CLIENT_RECT = window.HTMLElement.prototype.getBoundingClientRect;

beforeAll(() => {
    window.HTMLElement.prototype.getBoundingClientRect = () =>
        ({
            height: 800,
            width: 500
        }) as DOMRect;
});

afterAll(() => {
    window.HTMLElement.prototype.getBoundingClientRect = DEFAULT_BOUNDING_CLIENT_RECT;
});
```

brc-dd:
Shouldn't `overscan` still be respected though? Something like this should've been better:

```ts
rangeExtractor({
  ...range,
  overscan,
  count: outerSize === 0 ? 0 : count,
})
```

piecyk:
> Shouldn't `overscan` still be respected though? Something like this would've been better:
> 
> ```ts
> rangeExtractor({
>   ...range,
>   overscan,
>   count: outerSize === 0 ? 0 : count,
> })
> ```

don't think so, wouldn't it be counterintuitive that virtualizer returns items here

brc-dd:
I'm probably wrong, but I always thought of overscan as some sort of guarantee that at least that much rows will be rendered/returned. The docs define overscan as "The number of items to render above and below the visible area.' Here if the container size is 0, the visible area will be 0. Technically 0 + say 10 overscan should still be 10. But yeah I guess this is a bit up to how one is interpreting things.

piecyk:
Yeah, for example, do we still have visible area if it's 0, also if virtualizer returns range of type `{ startIndex: number; endIndex: number }` what it should return when size it's 0? This is used in rangeExtractor to calculate indexes.

brc-dd:
Ah, yeah. I guess this is fine. I'll create a separate issue if something goes south with this 👍

PR #603, opened 2023-10-08
fix: calculate range when outer size is defined and count > 0

### LINKED ISSUE #599: [v3.0.0-beta.63] Items not displayed in SSR with "useWindowVirtualizer"
### Describe the bug

On the version : v3.0.0-beta.63
When we SSR (Next or Remix) a page with "useWindowVirtualizer", the items do not display (because not returned by "getVirtualItems()") before we scroll.

If we can't scroll (because page is small) then the items never display.

https://codesandbox.io/s/new

### Steps to reproduce

1. Install version
2. Make a page in SSR with Next or Remix
3. Open the page

### Expected behavior

Items display as in .62

Every time

### Screenshots or Videos

_No response_

### Platform

Windows, Chrome

### tanstack-virtual version

v3.0.0-beta.63

### TypeScript version

v4

### Additional context

_No response_

### Terms & Code of Conduct

piecyk:
closes #599

### DIFF
--- a/packages/react-virtual/__tests__/index.test.tsx
+++ b/packages/react-virtual/__tests__/index.test.tsx
@@ -41,6 +41,10 @@ function List({
 
   const parentRef = React.useRef<HTMLDivElement>(null)
 
+  const elementRectCallbackRef = React.useRef<
+    ((rect: { height: number; width: number }) => void) | null
+  >(null)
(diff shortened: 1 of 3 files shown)


charkour:
Thanks for the quick fix!

Issue #654, opened 2024-01-26
Weird order of virtual items for many lanes when items has the same size


karolzeluk:
### Describe the bug

In case of many lanes defined and items having the same size, items might be ordered in weird way as they are sorted only by size.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/epic-bhaskara-fddtl2

### Steps to reproduce

Check how items are sorted in 2nd and 4th rows. They are sorted from left to right.

### Expected behavior

In all rows in case of the same size, items should be sorted from left to right.

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://github.com/TanStack/virtual/assets/77278511/5669f7f0-8edd-4ef8-a37b-60c5a0021b5d)

### Platform

- OS: macOS
- browser: Chrome
- version: 

### tanstack-virtual version

3.0.0-beta.54

### TypeScript version

v5.3.3

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

PR #655, opened 2024-01-26
fix: sort items by index in case they end in the same position

karolzeluk:
fixing: https://github.com/TanStack/virtual/issues/654


piecyk:
Thanks!

PR #670, opened 2024-02-17
feat: migrate to @tanstack/config

piecyk:


lachlancollins:
@piecyk it was easier to add _all_ the workspace config from the other projects than just a small piece of it, because of annoying TS/ESM/CJS/Nx issues. Aside from prettier formatting, no actual code has changed.

Changes:
- Replaced rollup with `@tanstack/config` build
- Replaced jest/babel with vitest
- Enabled proper Nx caching and modular tasks

PR #689, opened 2024-03-17
feat: add gap option to allow you to set the gap between items

gronxb:
## Motivation

Virtual lists operate on an `absolute` positioning basis, meaning that traditional list styling like `marginBottom` or `marginTop` doesn't work as expected. These styles require manual adjustment for proper spacing between items. Therefore, by introducing the gap option, we provide a convenient way to easily set spacing within the list.

https://github.com/TanStack/virtual/assets/41789633/b86d5e72-00d8-497c-b405-72ae5649a98f


piecyk:
Yeah let's merge it, saw a request for this few times already. Thank you for your contribution!

PR #708, opened 2024-04-16
feat: shouldAdjustScrollPositionOnItemSizeChange can be used from the…

Dima-Dim:
It is now impossible to use shouldAdjustScrollPositionOnItemSizeChange from the useVirtualizer and useWindowVirtualizer options And PR moves shouldAdjustScrollPositionOnItemSizeChange to options and can be passed when calling useVirtualizer and useWindowVirtualizer.

### DIFF
--- a/docs/api/virtualizer.md
+++ b/docs/api/virtualizer.md
@@ -320,6 +320,8 @@ scrollRect: Rect
 
 Current `Rect` of the scroll element.
 
+### `shouldAdjustScrollPositionOnItemSizeChange`
+
 ```tsx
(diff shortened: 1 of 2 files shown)


piecyk:
Thanks for the PR, it wa intentionally added as instance method, to alter the logic in event handlers. To always set it just assign it in place 


```js
const virtualizer = useVirtualizer({})
virtualizer.shouldAdjustScrollPositionOnItemSizeChange = () => false
```

piecyk:
> > Thanks for the PR, it wa intentionally added as instance method, to alter the logic in event handlers. To always set it just assign it in place
> 
> Okay, but wouldn't it be more convenient to pass the value of this method through options?

It depends, if you want to reset it back to undefined after some logic then no

Issue #715, opened 2024-04-25
Infinite loop in Vue adapter when estimated size doesn't match dynamic size


markjaniczak:
### Describe the bug

I'm using @tanstack/vue-virtual to render a dynamic list of elements and on specific browsers at specific resolution scaling virtualizer will end up in an infinite loop. I've found these previous issues which seem somewhat related:

https://github.com/TanStack/virtual/issues/619#issuecomment-1969516091
I put my measure function inside of a `nextTick` as to prevent that same issue.

https://github.com/TanStack/virtual/issues/159#issuecomment-873995201
I found the same issue on my windows machine at 125% scaling; what should be integer pixel sizes are read as decimal pixel sizes when calling `getBoundingClientRect` inside of my measure function.

In my actual app my list item elements will typically be one of two sizes as they conditionally render a line of text if it's available in the item. My codesandbox example is not exactly the same scenario as I've not been able to reproduce sub pixel sizes exactly but I've attached a couple of videos to help show what's happening.

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/runtime-river-jw2kt4

### Steps to reproduce

1. Load the page
2. Observe that the page is frozen in an infinite loop

### Expected behavior

As a user I was expecting the virtualizer to behave the same way on all screens regardless of pixel density.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/TanStack/virtual/assets/13617683/aadf7fd5-a1a8-4fb9-86f5-9c9924914277

### Platform

Windows 11 Pro 23H2
Edge 123.0.2420.97

### tanstack-virtual version

3.4.0

### TypeScript version

5.2.2

### Additional context

This issue was first noticed on a chromebook screen that had resolution scaling set to 125%. When it was changed to 100% the problem no longer appeared. On the same machine, I could move the window to a connected monitor that was set to 1920x1080 at 100% scaling and couldn't reproduce the issue either.

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

piecyk:
hmm @markjaniczak do you use some custom measureElement method, default one already use Math.round for exactly this reason https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts#L197-L216

markjaniczak:
[...]

Issue #619, opened 2023-11-03
Cannot scroll up in vue dynamic example


frkam:
### Describe the bug

Cannot scroll up in vue dynamic example

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/github/tanstack/virtual/tree/beta/examples/vue/dynamic?embed=1

### Steps to reproduce

1. Scroll down
2. Try to scroll up

### Expected behavior

I expect that it will scroll up without lag

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/TanStack/virtual/assets/80841256/093069c2-b270-4bf9-ad78-af488cdd666d

### Platform

OS - macos 17
Browser - chrome Version 118.0.5993.117 (Official Build) (arm64)

### tanstack-virtual version

See in codesandbox (ノ>_<)ノ

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

ssp7:
Any update on this?

KevinVandy:
> Any update on this?

Are you using Vue or React?

ssp7:
> > Any update on this?
> 
> Are you using Vue or React?

React

nconfrey:
I fixed a similar problem by modifying the Virtualizer to not remeasure elements when scrolling up:
```
measureElement: (element, entry, instance) => {
      const direction = instance.scrollDirection
      if (direction === "forward" || direction === null) {
        return element.scrollHeight
      } else {
        // don't remeasure if we are scrolling up
        const indexKey = Number(element.getAttribute("data-index"))
        let cacheMeasurement = instance.itemSizeCache.get(indexKey)
        return cacheMeasurement
      }
    }
 ```

**Comment by BazanAlexandro** [comment]
Was able to fix this delaying measurements with `nextTick` before measureElement. Looks like vue renders items a bit later, breaking dynamic measurements.

So, my measure function looks like this
```ts
  const measureElement = (el: any) => {
    nextTick(() => {
      if (!el) return;
      rowVirtualizer.value.measureElement(el);
    });
  };
```

AlejandroAkbal:
Thanks @BazanAlexandro, your solution works perfectly!

mvznfcoqe:
[...]

PR #716, opened 2024-04-26
Run watcher post DOM update in Vue adapter

markjaniczak:
This solves https://github.com/TanStack/virtual/issues/715

According to the [Vue docs](https://vuejs.org/guide/essentials/watchers.html#callback-flush-timing); watchers will run their effects after state updates but before DOM updates. The `ref` attribute performs the same way which is why some users discovered they need to wrap their `measureElement` calls with `nextTick` in order to measure elements in a post update state (https://github.com/TanStack/virtual/issues/619).

Currently the vue adapter will update state pre DOM update which causes an infinite loop if the estimated size of an element doesn't exactly match the measured size. Simply running the effect post DOM update will ensure that the elements are measured and taken into account before virtualizer updates its state.

### DIFF
--- a/packages/vue-virtual/src/index.ts
+++ b/packages/vue-virtual/src/index.ts
@@ -61,7 +61,7 @@ function useVirtualizerBase<
     },
     {
       immediate: true,
-      flush: 'post'
+      flush: 'post',
     },
   )
 
--- a/packages/vue-virtual/src/index.ts
+++ b/packages/vue-virtual/src/index.ts
@@ -61,6 +61,7 @@ function useVirtualizerBase<
     },
     {
       immediate: true,
+      flush: 'post'
     },
   )
 


piecyk:
@markjaniczak great thanks 👏 btw not that familiar with vue, why we need the first watch in adapter, this part https://github.com/TanStack/virtual/blob/a42c49afd8254365dbe4d36402b99cc1f5b455bf/packages/vue-virtual/src/index.ts#L36-L46

isn't it redundant as second one is watch whole options 🤔

piecyk:
@markjaniczak can you run formatting with `pnpm prettier:write`

markjaniczak:
> @markjaniczak great thanks 👏 btw not that familiar with vue, why we need the first watch in adapter, this part
> 
> https://github.com/TanStack/virtual/blob/a42c49afd8254365dbe4d36402b99cc1f5b455bf/packages/vue-virtual/src/index.ts#L36-L46
> 
> isn't it redundant as second one is watch whole options 🤔

I'm not very familiar with Vue either; I'm usually a React guy.

That said, I believe both are still necessary. The first watcher will react to changes of what `getScrollElement` returns and the second watcher will react to changes of the function's signature (if the function was saved as a vue ref).

PR #717, opened 2024-04-27
Don't run state effects in Vue immediately

markjaniczak:
Also related to https://github.com/TanStack/virtual/issues/715.

The previous PR https://github.com/TanStack/virtual/pull/716 partially solved the issue by not entering an infinite loop if the elements resized after the virtualizer had mounted but the issue persisted if the virtualizer had just mounted and the effect was running for the first time.

Confirmed in the codesandbox attached to the issue and my own project that this solves the issue.

### DIFF
--- a/packages/vue-virtual/src/index.ts
+++ b/packages/vue-virtual/src/index.ts
@@ -61,7 +61,7 @@ function useVirtualizerBase<
     },
     {
       immediate: true,
-      flush: 'post'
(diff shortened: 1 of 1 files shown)


markjaniczak:
These changes prevent an infinite loop if the elements are initially measured as a different size to their estimates but still enters one when the elements resize. @piecyk any idea where that might be happening?

PR #439, opened 2022-12-03
refactor: scroll handling

piecyk:
* smooth scroll not enabled by default
* drop enableSmoothScroll
* replace canSmooth with scroll behavior
* move adjustments logic in dynamic case to scroll method

PR #720, opened 2024-05-02
docs(virtualizer): update scrollToFn/Offset/Index options

jeffreykwan:
In https://github.com/TanStack/virtual/pull/439 (possibly previous commits)`scrollTo*` options argument was changed but docs weren't updated.


piecyk:
Thanks!

PR #724, opened 2024-05-05
feat: add enabled option to Virtualizer

sxriff:
[Related proposal](https://github.com/TanStack/virtual/discussions/722)

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -318,6 +318,7 @@ export interface VirtualizerOptions<
   initialMeasurementsCache?: VirtualItem[]
   lanes?: number
   isScrollingResetDelay?: number
+  enabled?: boolean
 }
 
 export class Virtualizer<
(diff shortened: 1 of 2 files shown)


KevinVandy:
I'm in favor of an enable/disable option. It would really help my MRT libraries.

packages/virtual-core/src/index.ts
```diff
@@ -328,6 +329,7 @@ export class Virtualizer<
   options!: Required<VirtualizerOptions<TScrollElement, TItemElement>>
   scrollElement: TScrollElement | null = null
   targetWindow: (Window & typeof globalThis) | null = null
+  private enabled: boolean = true
```
piecyk, line 332: I think we don't need an instance variable, overall we should relay on scrollElement if it's null then everything should be reseted.

enabled={false} should be same as getScrollElement: () => null

piecyk:
@sxriff after reading the PR refactored it bit to handle reseting scroll offset, measurementsCache etc it's bit tricky when we also want to handle the initialRect, initialOffset etc, checkout the #741

piecyk:
Thanks again for the PR @sxriff! We ended up merging another one #741 that covered this.

Closing this now. Appreciate your efforts and look forward to more contributions!

packages/virtual-core/src/index.ts
sxriff: Yes, I guess that's a correct flow

sxriff:
> Thanks again for the PR @sxriff! We ended up merging another one #741 that covered this.
> 
> Closing this now. Appreciate your efforts and look forward to more contributions!

@piecyk Thank you for reviewing and adding the feature, looking forward to using it in my project!

PR #737, opened 2024-06-01
feat: add measureElement on virtual item, when we can't relay on node.isConnected

piecyk:
As we can relay on node.isConnected to attach the RO or read size, like in `lit` integration https://github.com/TanStack/virtual/pull/718 we need to have another option to do. In this PR  https://github.com/TanStack/virtual/pull/737 let's add back the measureElement on virtual item and keep it stale via measureElementCache that has bound in scope index.

```
${ref(virtualRow.measureElement}
```


kkx64:
@piecyk @tannerlinsley  This is all good and well, but changing the type of `VirtualItem` to `VirtualItem<T>` in a minor release made our UI component builds randomly fail. Please be careful of these types of changes.

PR #738, opened 2024-06-04
fix(virtual-core): add support to multi window

YuanboXue-Amber:
## Overview

In multi window environment, where React is running in a main window/process and renders content to a different child window, the globals like `setTimeout`/`clearTimeout` will not work properly because it's using the main Window object and not the child one.

This PR adds support to multi-window.
`setTimeout` --> `targetWindow.setTimeout`
`clearTimeout` --> `targetWindow.clearTimeout`
`ResizeObserver` --> `targetWindow.ResizeObserver`

`targetWindow` is obtained from `scrollElement.ownerDocument.defaultView`. 

## More details
We have multi-windows use case, so we want to make sure that the globals work well with it. This is the reason for my PR.

I built an example of dynamic height list in child window:
https://stackblitz.com/edit/vitejs-vite-2zau9t?file=src%2FApp.tsx&terminal=dev
1. Open this example on Chrome/Edge
2. Popout the example in a new tab
3. Click 'Open Child Window' button, a child window opens and a dynamic height virtual list is visible
4. Open devtools on child window, modify the content of Row 0. 
5. Notice the resize did not get triggered immediately. It triggers after 5-10 seconds. 
(Note that this happens only when the child window is opened for the first time from the main window. This inconsistent behavior came from browser.)

I tried the same example with my change, and the resize is triggered immediately.

- Before

https://github.com/TanStack/virtual/assets/28751745/8831ecaf-2633-40a0-8244-2e40ac8044d7

- After

https://github.com/TanStack/virtual/assets/28751745/2871cb72-fd11-432f-9c3a-fd6974e659f1

>We are switching to @tanstack/react-virtual for our virtualized lists because we find its api very appealing! 
This is my first time contributing to tanstack. I appreciate your feedbacks!

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -353,7 +353,7 @@ export class Virtualizer<
         return _ro
       }
 
-      if (typeof ResizeObserver === 'undefined' || !this.targetWindow) {
+      if (!this.targetWindow || !this.targetWindow.ResizeObserver) {
         return null
       }
 
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -67,6 +67,10 @@ export const observeElementRect = <T extends Element>(
   if (!element) {
     return
   }
+  const targetWindow = element.ownerDocument.defaultView
+  if (!targetWindow) {
+    return
+  }
 
   const handler = (rect: Rect) => {
     const { width, height } = rect
@@ -79,7 +83,7 @@ export const observeElementRect = <T extends Element>(
     return () => {}
   }
 
-  const observer = new ResizeObserver((entries) => {
+  const observer = new targetWindow.ResizeObserver((entries) => {
     const entry = entries[0]
     if (entry?.borderBoxSize) {
       const box = entry.borderBoxSize[0]
@@ -134,13 +138,21 @@ export const observeElementOffset = <T extends Element>(
   if (!element) {
     return
   }
+  const targetWindow = element.ownerDocument.defaultView
+  if (!targetWindow) {
+    return
+  }
 
   let offset = 0
   const fallback = supportsScrollend
     ? () => undefined
-    : debounce(() => {
-        cb(offset, false)
-      }, instance.options.isScrollingResetDelay)
+    : debounce(
+        targetWindow,
+        () => {
+          cb(offset, false)
+        },
+        instance.options.isScrollingResetDelay,
+      )
 
   const createHandler = (isScrolling: boolean) => () => {
     offset = element[instance.options.horizontal ? 'scrollLeft' : 'scrollTop']
@@ -168,13 +180,21 @@ export const observeWindowOffset = (
   if (!element) {
     return
   }
+  const targetWindow = element.window
+  if (!targetWindow) {
+    return
+  }
 
   let offset = 0
   const fallback = supportsScrollend
     ? () => undefined
-    : debounce(() => {
-        cb(offset, false)
-      }, instance.options.isScrollingResetDelay)
+    : debounce(
+        targetWindow,
+        () => {
+          cb(offset, false)
+        },
+        instance.options.isScrollingResetDelay,
+      )
 
   const createHandler = (isScrolling: boolean) => () => {
     offset = element[instance.options.horizontal ? 'scrollX' : 'scrollY']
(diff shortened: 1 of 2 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -330,15 +351,17 @@ export class Virtualizer<
...
-      } else if (typeof ResizeObserver !== 'undefined') {
-        return (_ro = new ResizeObserver((entries) => {
-          entries.forEach((entry) => {
-            this._measureElement(entry.target as TItemElement, entry)
-          })
-        }))
-      } else {
+      }
+
+      if (typeof ResizeObserver === 'undefined' || !this.targetWindow) {
```
piecyk, line 356: I think here we would like to check if ResizeObserver is on targetWindow?

packages/virtual-core/src/index.ts
```diff
@@ -79,7 +83,7 @@ export const observeElementRect = <T extends Element>(
     return () => {}
   }
 
-  const observer = new ResizeObserver((entries) => {
+  const observer = new targetWindow.ResizeObserver((entries) => {
```
piecyk, line 86: Also check if ResizeObserver is on targetWindow

packages/virtual-core/src/index.ts
YuanboXue-Amber: Do you mean a check like:
```tsx
      if (
        typeof ResizeObserver === 'undefined' ||
        !this.targetWindow ||
        !this.targetWindow.ResizeObserver
      ) {
        return null
      }
```
I think it's not necessary because this should be guarded by type. When used in SSR `this.targetWindows` (`this.scrollElement.ownerDocument.defaultView`) should be null

packages/virtual-core/src/index.ts
```diff
@@ -307,8 +327,9 @@ export class Virtualizer<
   private unsubs: (void | (() => void))[] = []
   options!: Required<VirtualizerOptions<TScrollElement, TItemElement>>
   scrollElement: TScrollElement | null = null
+  private targetWindow: (Window & typeof globalThis) | null = null
```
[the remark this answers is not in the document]
YuanboXue-Amber: Sounds good. I just wasn't sure about changing the public API so I went with private option. Please let me know if you prefer public targetWindow option :)

packages/virtual-core/src/index.ts
```diff
@@ -67,6 +67,10 @@ export const observeElementRect = <T extends Element>(
   if (!element) {
     return
   }
+  const targetWindow = element.ownerDocument.defaultView
```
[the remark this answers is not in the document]
YuanboXue-Amber: Updated :) (Together with an update on `Virtualizer` class to change `targetWindow` from private to public)

packages/virtual-core/src/index.ts
YuanboXue-Amber: Here it should be guarded by type as well. (If you meant by checking `if (targetWindow.ResizeObserver)`)

packages/virtual-core/src/index.ts
piecyk: No, imho we don't need to check ResizeObserver on global context if we read it from this.targetWindow

```tsx
if (!this.targetWindow || !this.targetWindow.ResizeObserver) {
  return null
}
```

packages/virtual-core/src/index.ts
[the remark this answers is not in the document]
YuanboXue-Amber: And if we're going to add a public targetWindow option, should it defaults to `this.scrollElement.ownerDocument.defaultView` when not provided

packages/virtual-core/src/index.ts
YuanboXue-Amber: Right! I updated it according to the suggestion

packages/virtual-core/src/index.ts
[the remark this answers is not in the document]
piecyk: On other side keeping it on instance should be good enough, so let's not add it to options just now.

packages/virtual-core/src/index.ts
piecyk: Thanks.

packages/virtual-core/src/index.ts
YuanboXue-Amber: Updated :) could you give it one more round of review

piecyk:
Thanks @YuanboXue-Amber great work 👏

YuanboXue-Amber:
@piecyk thanks for the quick review! When do you plan the next release? We would like to use have this change in our app.

piecyk:
@YuanboXue-Amber it alredy done https://github.com/TanStack/virtual/releases/tag/v3.5.1

PR #774, opened 2024-07-25
docs: update API documentation of scrollToIndex

fredericbahr:
The interface defines that the scroll behavior of the `scrollToIndex` function can be 'auto' or 'smooth'.

The documentation falsly stated 'auto' or 'scroll'

PR #775, opened 2024-07-25
replace 'scroll' with 'smooth' for scrollToIndex

mkaraula:
Just realized there is a slight error in the Docs. 'smooth' is the actual option and not 'scroll'.


mkaraula:
Closed as I have just seen someone already opened a PR: https://github.com/TanStack/virtual/pull/774

PR #779, opened 2024-07-29
fix(examples): Remove conflicting React type definitions

SaeWooKKang:
## What this PR solves
- It solves the `jsx` type error problem in certain examples of official documents.
스크린샷 2024-07-29 오후 6 16 28

### Type Error List 
- [fixed](https://tanstack.com/virtual/latest/docs/framework/react/examples/fixed)
- [dynamic](https://tanstack.com/virtual/latest/docs/framework/react/examples/dynamic)
- [table](https://tanstack.com/virtual/latest/docs/framework/react/examples/table)
- [window](https://tanstack.com/virtual/latest/docs/framework/react/examples/window)

## Summary
- I think the problem is that the type definition is duplicated.

## Steps to Reproduce
1. Go to the example project page of the official document created above.
2. The package goes down and waits to run.
3. Check the type error in the `jsx` part.

## How to Test the Solution
``` 
git clone https://github.com/TanStack/tanstack.com.git

cd ./tanstack.com

Go to the file 
- app/routes/$libraryId.$version.docs.framework.$framework.examples.$.tsx

Change the value of the 'stackBlitzUrl' variable to 
  'https://stackblitz.com/github/SaeWooKKang/virtual/tree/main/examples/react/fixed?embed=1&theme=light&preset=node&file=src/main.tsx'.

pnpm install

pnpm dev

Check example in browser
```


piecyk (changes requested):
@SaeWooKKang i think the issue is not with react type definitions, but with typescript configuration, we should replace tsconfig.dev.json with tsconfig.json. We can follow what https://github.com/TanStack/query is doing.

SaeWooKKang:
@piecyk Thank you for your quick feedback.
As you said, I corrected it by referring to the example and confirmed that it worked properly. Should I correct other examples that do not have errors for consistency?

examples/react/dynamic/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true
+  },
+  "files": ["src/main.tsx"],
+  "include": [
+    "src"
+    // "__tests__/**/*.test.*"
+  ]
```
piecyk, line 28: We don't need files property, just include src would be enough
```suggestion
   "include": ["src"]
```

piecyk (changes requested):

piecyk:
> @piecyk Thank you for your quick feedback. As you said, I corrected it by referring to the example and confirmed that it worked properly. Should I correct other examples that do not have errors for consistency?

Yes, please 🙌

SaeWooKKang:
> @piecyk Thank you for your quick feedback. As you said, I corrected it by referring to the example and confirmed that it worked properly. Should I correct other examples that do not have errors for consistency?
Yes, please 🙌

@piecyk When I set the tsconfig.json to `strict: true`, there are errors in components that have not been interface-defined. How should I fix it?
1. set `strict:false`, modify it in another pr
2. set `strict: true`, define interface

piecyk (approved):

PR #787, opened 2024-08-07
feat: Support inverted horizontal scrolling for right-to-left locales

cachgill:
This change adds an option for consumers who need to switch between left-to-right and right-to-left language locales. 

With this change, scrolling left-to-right:
https://github.com/user-attachments/assets/93892f15-a5fe-4cdc-8ccf-45c316a966f0

Without this change, scrolling right-to-left:
https://github.com/user-attachments/assets/47aa1cae-de3d-4946-9fd8-29fee19fe560

With this change, scrolling right-to-left:
https://github.com/user-attachments/assets/ab87e0f7-0679-4669-986d-69188d572f7f

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -156,7 +156,10 @@ export const observeElementOffset = <T extends Element>(
       )
 
   const createHandler = (isScrolling: boolean) => () => {
-    offset = element[instance.options.horizontal ? 'scrollLeft' : 'scrollTop']
+    const { horizontal, isRtl } = instance.options
+    offset = horizontal
+      ? element['scrollLeft'] * (isRtl && -1 || 1)
+      : element['scrollTop']
     fallback()
     cb(offset, isScrolling)
   }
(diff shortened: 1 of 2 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -156,7 +156,10 @@ export const observeElementOffset = <T extends Element>(
       )
 
   const createHandler = (isScrolling: boolean) => () => {
-    offset = element[instance.options.horizontal ? 'scrollLeft' : 'scrollTop']
+    const { horizontal, isRtl } = instance.options
+    offset = horizontal
+      ? element['scrollLeft'] * (isRtl && -1 || 1)
```
piecyk, line 161: Let's keep it simple 
```suggestion
      ? element['scrollLeft'] * (isRtl ? -1 : 1)
```

Did you test it on all modern browsers?

cachgill: We are used to adhering to ESlint's rules regarding ternaries: https://eslint.org/docs/latest/rules/no-nested-ternary

The scenario in the included videos was tested in Safari, Chrome, Firefox, Brave, Opera, and Edge.

piecyk:
Thanks @cachgill 🙌

PR #806, opened 2024-08-21
refactor: memoize notify

piecyk:


andreial (approved):
👍

guysympoly:
Hi @piecyk, please note that removing `measureElement` and the generic signature `<TItemElement extends Element>`  from `VirtualItem` is a breaking change. It just bit me really hard during a pipeline.

piecyk:
> Hi @piecyk, please note that removing `measureElement` and the generic signature `<TItemElement extends Element>` from `VirtualItem` is a breaking change. It just bit me really hard during a pipeline.

Yes, sorry for that, got mixed up with auto publish script we have.

juzerzarif:
Will there be a release on the v3 line to fix the breaking change? Also just ran into this

piecyk:
> Will there be a release on the v3 line to fix the breaking change? Also just ran into this

Yeah sorry for that, not planing any more on 3x. 

In the end adding generic to VirtualItem was a breaking change at first.

PR #814, opened 2024-08-28
fix(virtual-core): add bigint to Key to align with @types/react implementation

huv1k:
This PR enhances `Key` type with `bigint` to align with `React.Key`. This was updated in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66723. This will enable us to use `React.Key` as type for where we require `Key` as type.

### DIFF
--- a/docs/api/virtual-item.md
+++ b/docs/api/virtual-item.md
@@ -6,7 +6,7 @@ The `VirtualItem` object represents a single item returned by the virtualizer. I
 
 ```tsx
 export interface VirtualItem {
-  key: string | number
+  key: string | number | bigint
   index: number
   start: number
   end: number
@@ -19,7 +19,7 @@ The following properties and methods are available on each VirtualItem object:
 ### `key`
 
 ```tsx
-key: string | number
+key: string | number | bigint
 ```
 
 The unique key for the item. By default this is the item index, but should be configured via the `getItemKey` Virtualizer option.
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -26,7 +26,7 @@ export interface Range {
   count: number
 }
 
-type Key = number | string
+type Key = number | string | bigint
 
 export interface VirtualItem {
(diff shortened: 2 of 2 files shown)


SaeWooKKang:
If changed like that, it might be more convenient when using React. 

However, Should the higher-level element 'core' depend on the lower-level element 'react'? 
How about using type casting in React?

``` ts
const alsoHuge = BigInt(9007199254740991);

alsoHuge.toString();
```

piecyk:
> However, Should the higher-level element 'core' depend on the lower-level element 'react'?

@SaeWooKKang allowing bigint as key is not really connected to react

huv1k:
Yes, the problem is when you use `Key` from react, because it has a different type and there is an error, which shouldn't be the case. Is there anything more needed?

SaeWooKKang:
> @SaeWooKKang allowing bigint as key is not really connected to react

You're right, there isn't a direct connection. However, my comment was about whether it's appropriate to add a type to the key specifically for the purpose of using it with React.

> Yes, the problem is when you use Key from react, because it has a different type and there is an error, which shouldn't be the case. Is there anything more needed?

Is there no need to modify the documentation for the type?

piecyk:
> However, my comment was about whether it's appropriate to add a type to the key specifically for the purpose of using it with React.

Yep, thanks for the input. I think we are fine here, as bigint is overall an primitive not a react thing.

> Is there no need to modify the documentation for the type?

good catch, @huv1k can you update docs at https://github.com/TanStack/virtual/blob/main/docs/api/virtual-item.md

kkx64:
Why not just use the React type for keys???

piecyk:
> Why not just use the React type for keys?

As core should not be bound to any lib, adding the bigint created other issues like #819

kkx64:
> > Why not just use the React type for keys?
> 
> As core should not be bound to any lib, adding the bigint created other issues like #819

Fair enough, I forgot that the core can be used across different frameworks

PR #816, opened 2024-08-28
fix: When using initialRect, the first rendered nodes do not observe ResizeObserver.

### LINKED ISSUE #815: When using initialRect, the first rendered nodes do not observe ResizeObserver.
### Describe the bug

When using initialRect, the first rendered nodes do not observe ResizeObserver.

Therefore, resizing the window will result in a broken layout.

This is because `measureElement` is called before `_willUpdate` is called and therefore `virutalizer.targetWindow` is `null` when `measureElement` is called.

https://codesandbox.io/p/sandbox/white-breeze-5vg7lf

### Steps to reproduce

1. Resize window.

### Expected behavior

I expect that the first rendered nodes will respond to resize.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/c93e25aa-efa1-48f0-a470-962b27457119

### Platform

- OS: macOS, Browser: Chrome,Safari

### tanstack-virtual version

3.10.5

### TypeScript version

4.4.4

### Additional context

This problem is solved by setting `virtualizer.targetWindow && virtualizer.measureElement` instead of `virtualizer.measureElement` in the `ref` as a workaround.

Or, it can be solved by calling `document.querySelectorAll(“data-index”).forEach(virtualizer.measureElement)` in `useEffect`.

### Terms & Code of Conduct

MukaNakazato:
This fix changes the reference to `instance.measureElement` so that the ref callback is called again after `instance._willUpdate()` is called.

This PR resolves #815 .

### DIFF
--- a/packages/react-virtual/src/index.tsx
+++ b/packages/react-virtual/src/index.tsx
@@ -47,7 +47,8 @@ function useVirtualizerBase<
   }, [])
 
   useIsomorphicLayoutEffect(() => {
-    return instance._willUpdate()
+    instance._willUpdate()
+    instance.measureElement = instance.measureElement.bind(instance)
   })
 
   return instance


packages/react-virtual/src/index.tsx
```diff
@@ -47,7 +47,8 @@ function useVirtualizerBase<
   }, [])
 
   useIsomorphicLayoutEffect(() => {
-    return instance._willUpdate()
+    instance._willUpdate()
+    instance.measureElement = instance.measureElement.bind(instance)
```
piecyk, line 51: Kinda feels like this is more core issue not react specific, hmm what about something like

As we should have elements in cache, we can observe elements when we set scroll element
 
Zrzut ekranu 2024-08-29 o 11 47 58

MukaNakazato: Hmm, since `this.cleanup()` gets called in `_willUpdate`, `this.elementsCache` ends up being empty... It seems necessary to call `cleanup` when `this.scrollElement` changes, and it feels natural to clear the `elementsCache` within `cleanup`, but I can't figure out what the best solution is...

How about making it event-driven and only calling _measureElement after _willUpdate has been called?
I thought so, but it would be better if cleanup() also handled event listeners.

Hmm,I can’t come up with a good idea.

piecyk: True, this is bit tricky as when clearing scrollElement we can set it back, but with elementsCache we can't. Kinda feels maybe just stop observing with observer.disconnect() and observe back. 

> How about making it event-driven and only calling _measureElement after _willUpdate has been called?

When we observe back elements then _measureElement will be called from RO callback.

piecyk: Maybe for now let's not clear the cache on clear and observe it back later https://github.com/TanStack/virtual/pull/830

Issue #750, opened 2024-07-05
HMR (live reload, fast refresh) broke render in React


dartess:
### Describe the bug

After the hot reload, new list items are not rendered

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-virtual-dc5j1h?file=src%2Fmain.tsx,src%2FRowVirtualizerFixed.tsx&preset=node

### Steps to reproduce

1. open repl
2. edit `console.log('foo');` with some different text
3. press cmd/ctrl+S for save and live reload
4. scroll list
5. see that list is empty

### Expected behavior

it should just work

### How often does this bug happen?

Every time

### Platform

any platform , any bundler — vite, wepack, rspack

### tanstack-virtual version

3.8.1

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

mr-devboy:
Same here

hophiphip:
TLDR: `scrollElement` gets cleared in `_didMount` and that is why virtualizer stops working. 



The only methods that can change `scrollElement` are: `_willUpdate` and `_didMount`. 

1. On the initial run `_willUpdate` and `_didMount` are called in this order:
![2024-08-29-16:45:41-screenshot](https://github.com/user-attachments/assets/aefcd879-67d2-4a8d-bd2c-5c82f2fbb2d5)

2. After triggering HMR  `_willUpdate` and `_didMount` will be called in this order:
![2024-08-29-16:48:18-screenshot](https://github.com/user-attachments/assets/92114f6d-9da9-4522-84e6-88c9fff2cd30)

[2] - This means `scrollElement` gets cleared and set to `null` which "breaks" the virtualizer.


You can debug this in this CodeSandbox: https://codesandbox.io/p/devbox/4wnzrm
You can also find fixed `useVirtualizerBase` in `src/hooks/useVirtualizerFixed.tsx`.


NOTE!: Im not 100% sure if this is a valid fix (but is seems to work in DEV and PROD, and tests pass), but I had to change 
`useVirtualizerBase` hook [L45-L47](https://github.com/TanStack/virtual/blob/90ede1e03c0314477056c99e6d646757ed2403d5/packages/react-virtual/src/index.tsx#L45) from:
```ts
React.useEffect(() => {
    return instance._didMount();
  }, []);
```

to:

```ts
React.useEffect(() => {
[...]

PR #820, opened 2024-08-30
fix(react-virtual): handle `scrollElement` cleanup after HMR

hophiphip:
When HMR is triggered `useEffect` cleanup function will call function returned from `instance._didMount()`.
This will set `scrollElement` to null and `Virtualizer` will stop working.

I added an additional check if `scrollElement` is null and if it is - call `instance._willUpdate()`.

This PR resolves https://github.com/TanStack/virtual/issues/750.

### DIFF
--- a/packages/react-virtual/src/index.tsx
+++ b/packages/react-virtual/src/index.tsx
@@ -42,7 +42,7 @@ function useVirtualizerBase<
 
   instance.setOptions(resolvedOptions)
 
-  React.useEffect(() => {
+  useIsomorphicLayoutEffect(() => {
     return instance._didMount()
   }, [])
 


Innei (approved):

rameshx:
@piecyk can this be merged? I think this fixes the white space issue after HMR

Innei:
@tannerlinsley Hey, can you take a look at this PR? Can it be merged?

packages/react-virtual/src/index.tsx
```diff
@@ -43,6 +43,10 @@ function useVirtualizerBase<
   instance.setOptions(resolvedOptions)
 
   React.useEffect(() => {
+    if (!instance.scrollElement) {
+      instance._willUpdate()
+    }
+
```
piecyk, line 49: I think better would be to change the `useEffect` to `useIsomorphicLayoutEffect`
```suggestion
  useIsomorphicLayoutEffect(() => {
```

then order will be correct 
```
instance._didMount, cleanup
instance._willUpdate, cleanup
instance._didMount
instance._willUpdate
```

piecyk (changes requested):

packages/react-virtual/src/index.tsx
hophiphip: Hmm, interesting, was there a reason `useEffect` was used instead of  `useIsomorphicLayoutEffect` since the [beginning](https://github.com/TanStack/virtual/commit/4f38da468e5db4f7760d2c7845bc7d624fdcf40d#diff-d7f76e6511ee16ef9493c403e9f81f738e6d281dd9490890be946a9811f73d7aR39)?

piecyk: Overall when possible useEffect should preferred, there was no clear reason till now ot use  `useIsomorphicLayoutEffect` imho. 

> useLayoutEffect is a version of [useEffect](https://react.dev/reference/react/useEffect) that fires before the browser repaints the screen.

As this is called in limited way imho changing it to layout effect should be fine.

piecyk: @hophiphip please update the PR, then we can merge it! Thanks for looking into this.

hophiphip: @piecyk, updated PR

piecyk (approved):

PR #834, opened 2024-09-10
docs(examples): fix bug in Vue infinite scroll example

BrianMitchL:
This fixes a ref not being accessed with it's `.value` accessor, causing a boolean check to always be truthy.


kouts (approved):

kouts:
Thanks for the fix @BrianMitchL

PR #842, opened 2024-09-25
Fix(virtual-core): Wrap ResizeObserver callbacks in window.requestAnimationFrame to avoid error

samstronghammer:
This PR wraps the two ResizeObserver callbacks with a window.requestAnimationFrame so that the "ResizeObserver loop completed with undelivered notifications" error doesn't happen.

Problem: ResizeObserver callbacks throw errors if they aren't able to resolve fully when they are invoked. This can happen if the callbacks are called at an inopportune time for the main thread, resulting in callbacks being dropped and the "ResizeObserver loop completed with undelivered notifications" error. This varies depending on how and when the resizes happened, and what browser you're on. I've found that Firefox is particularly picky about this.

Solution: There are several suggested [here](https://stackoverflow.com/questions/76187282/react-resizeobserver-loop-completed-with-undelivered-notifications). In my personal experience, window.requestAnimationFrame solves the problem pretty well. Wrapping the two callbacks in the ResizeObservers here resulted in the errors completely stopping for my use case.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -84,15 +84,17 @@ export const observeElementRect = <T extends Element>(
   }
 
   const observer = new targetWindow.ResizeObserver((entries) => {
-    const entry = entries[0]
-    if (entry?.borderBoxSize) {
-      const box = entry.borderBoxSize[0]
-      if (box) {
-        handler({ width: box.inlineSize, height: box.blockSize })
-        return
+    window.requestAnimationFrame(() => {
+      const entry = entries[0]
+      if (entry?.borderBoxSize) {
+        const box = entry.borderBoxSize[0]
+        if (box) {
+          handler({ width: box.inlineSize, height: box.blockSize })
+          return
+        }
       }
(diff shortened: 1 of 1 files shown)


piecyk:
@samstronghammer most cases this error has no real effect, other than log in console what is annoying. Wrapping it with requestAnimationFrame would creates it's own issues.

piecyk:
Thanks for sharing more details, and I totally understand the impact this has on your workflow. I’ll try to debug the issue further and see if I can provide a workaround to minimize the problem, or use requestAnimationFrame, let see. I’ll keep you updated once I have more insights or a better approach.

olavgu:
@piecyk Any updates on this? I'm experiencing the same issue as @samstronghammer

HodgeWen:
After updating the version of Edge, this issue also occurred.

piecyk:
Closing this PR in favor of #923 that includes the useAnimationFrameWithResizeObserver option to control if ResizeObserver should be wrapped with AnimationFrame

PR #851, opened 2024-10-10
feat(react-virtual): compatibility with the React compiler

### LINKED ISSUE #736: React Compiler Compat: useVirtual results are cached at their initial values
### Describe the bug

An usage of `useVirtual` gets compiled by React Compiler into something like this:

```
const virtualizer = (0,
_tanstack_react_virtual__WEBPACK_IMPORTED_MODULE_9__.useVirtualizer)(t3);
let t4;
if ($[9] !== virtualizer) {
  t4 = virtualizer.getVirtualItems();
  $[9] = virtualizer;
  $[10] = t4;
} else {
  t4 = $[10];
}
```

Since the virtualizer ref never changes, the virtual items never get updated.

Not sure if this should be a Tanstack Virtual issue or React Compiler needs to apply less assumptions when caching values, but thought that would be good to flag the problem here.

Wrapping `useVirtual` into something like this fixes the issue:

```js
const useVirtualWrapper = (...args) => {
  'use no memo';
  return { ...useVirtual(...args) };
}
```

https://codesandbox.io/p/devbox/runtime-dew-9rlvm7

### Steps to reproduce

1. Open the example (forked from https://tanstack.com/virtual/latest/docs/framework/react/examples/fixed)
2. The virtual lists are rendering no items

### Expected behavior

The UI should be the same as https://tanstack.com/virtual/latest/docs/framework/react/examples/fixed

Every time

### Screenshots or Videos

_No response_

### Platform

- MacOS, Chrome and Firefox

### tanstack-virtual version

3.5.0

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

### LINKED ISSUE #743: React 19 issue with useWindowVirtualizer
### Describe the bug

Sorry for the drive by issue… 

We ran into a problem with react 19 today that was a pita to track down. So dropping it here incase it's useful to anyone.

We had a simple useWindowVirutalizer in a client component in next.js w/ react 19 (like this):

```js
const virtualizer = useWindowVirtualizer({
		count: branches.length,
		estimateSize: () => 66,
		overscan: 5,
		scrollMargin: listRef.current?.offsetTop ?? 0,
	});
```

and virtualizerRef.getVirtualItems() was being hoisted / memo'd by the compiler in such a way that it was always returning 0 for `getVirutalItems`.

To fix it we just dropped the virtualizer in a ref…

```js
	const virtualizer = useWindowVirtualizer({
		count: branches.length,
		estimateSize: () => 66,
		overscan: 5,
		scrollMargin: listRef.current?.offsetTop ?? 0,
	});

	// we do this because react 19 was optimzing it away
	const virtualizerRef = useRef(virtualizer);
	
	…
	
	virtualizerRef.current.getVirtualItems() // etc…
```

n/a

### Steps to reproduce

Don't have time to spin up a test case sorry! 

### Expected behavior

Expected getVirtualItems() to return items.

Every time

### Screenshots or Videos

_No response_

### Platform

macos, chrome, react 19

### tanstack-virtual version

3.7.0

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

gdorsi:
Fixes #736 
Fixes #743 

This PR addresses a critical issue with useVirtualizer compatibility when used alongside the React compiler. Previously, the React compiler would cache results based on the virtualizer instance, preventing updates when virtual items changed.

The solution implements a wrapper using Object.create around the virtualizer instance. 

This approach maintains the internal state while allowing the instance reference to change. 

The internal state remains constant across updates, with only the instance reference being modified when virtual items change.

This change affects object property behavior. Spread operations and Object.keys() on the virtualizer now return empty results. This impacts code relying on these operations and could be considered a breaking change.

If this, or the `getVirtualItems` eager evaluation, is a cause of concern we could make the transformation opt-in via option flag or could even be implemented externally as a separate hook.


packages/react-virtual/src/index.tsx
```diff
@@ -50,7 +50,7 @@ function useVirtualizerBase<
     return instance._willUpdate()
   })
 
-  return instance
+  return React.useMemo(() => Object.create(instance), [instance.getVirtualItems()])
```
gdorsi, line 53: We could move this function inside the virtual core and use the `memo` helper to invalidate the instance generation.

Should give us more flexibility and will be easier to cover more invalidations.

BjoernRave:
lookinf forward for this to get merged, especially since with react 19 rc1 things seem to start moving again

piecyk:
Thinking about this, if we want to move in this direction, useSyncExternalStore feels like a more natural fit compared to useMemo. We can make it as separate hook"

For now 'use no memo' seems like a reasonable temporary workaround! https://react.dev/learn/react-compiler#something-is-not-working-after-compilation

gdorsi:
@piecyk 

Would somthing like this work for you?

```ts
function useMemoSafeVirtualizer<
  TScrollElement extends Element | Window,
  TItemElement extends Element,
>(
): Virtualizer<TScrollElement, TItemElement> {
  const [instance] = React.useState(
    () => new Virtualizer<TScrollElement, TItemElement>(options),
  )

  instance.setOptions(options)

  React.useEffect(() => {
    return instance._didMount()
  }, [])

    return instance._willUpdate()
  })

  return useSyncExternalStore(
    useCallback((onStoreChange) => instance.subscribe(onStoreChange), [instance]),
    () => instance.getMemoSafeInstance(),
    () => instance.getMemoSafeInstance(),
  )
}
```

We would need to add subscribe and getMemoSafeInstance to the core:
```ts
  getMemoSafeInstance = memo(
    () => [this.getVirtualItems()],
    () => Object.create(this),
    {
      key: process.env.NODE_ENV !== 'production' && 'getMemoSafeInstance',
      debug: () => this.options.debug,
    },
  )
   
  private notify = (sync: boolean) => {
    this.options.onChange?.(this, sync)
    this.subscribers.forEach(fn => fn(this, sync));
  }
  
  subscribe = (callback) => {
    this.subscribers.add(callback)

    retrurn () => {
        this.subscribers.delete(callback)
    }
  }
```

The-Best-Codes:
Please merge this, it would be awesome!

samesfahani-tuplehealth:
Any timeline on this PR getting merged?

gdorsi:
This is not really a fully featured fix, more like a POC.

Going to close to avoid unnecessary distractions.

PR #862, opened 2024-10-25
docs: extend scrollMargin description

Wordllban:
As discussed here https://github.com/TanStack/virtual/discussions/856#discussioncomment-10955571 I would like to update the docs for the scrollMargin argument, so other people will understand better how to properly use it


piecyk:
@Wordllban thanks!

PR #863, opened 2024-10-26
chore(deps): update svelte-virtual peerDependencies to allow for v5

nigel-dev:
Svelte v5 was just released and is backward compatible with v4. This updates the peerDependencies to allow for those running Svelte 5 projects to pull in this dependency.

For reference:
https://svelte.dev/blog/svelte-5-is-alive


wobsoriano (commented):
I believe it's safe to remove v3 as peer dep. What do you think?

nigel-dev:
> I believe it's safe to remove v3 as peer dep. What do you think?

I can do that, just didnt want to break anyone who might be still using version 3.

piecyk:
> > I believe it's safe to remove v3 as peer dep. What do you think?
> 
> I can do that, just didnt want to break anyone who might be still using version 3.

that is good question, maybe let's leave the v3 for now.

nigel-dev:
@piecyk So add v3 support back?

piecyk:
> @piecyk @wobsoriano
> 
> So add v3 support back?

Yep if it works with v3 don't see any reason to remove it.

Issue #799, opened 2024-08-17
Additional scroll events after scrollend causes isScrolling to remain true


martinpengellyphillips:
### Describe the bug

Additional `scroll` events sometimes occur _after_ `scrollend` has been fired. This then causes `isScrolling` to be set to `true` indefinitely (until new scrolling is performed). 

It is unclear if this is specific to Solid adapter, but I did confirm the events are received in the wrong order in the default `observeElementOffset` handler. 

One could argue this is a browser implementation issue, but it might still be worth investigating more and potentially defending against by always setting `isScrolling` to false a short time after last scroll event received 🤔 

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/solidjs-virtual-isscrolling-bug-r75q6y?file=%2Fsrc%2Fmain.tsx%3A7%2C26

### Steps to reproduce

1. Add a conditional rendering of placeholder rows in a virtual list (based on `isScrolling`).
2. Scroll up and down and then stop scrolling*
3. Sometimes the placeholder rows will not be replaced on scroll end because `isScrolling` is still `true`.

* On Firefox, using the keyboard to scroll can often trigger the issue. On Chrome, scrolling rapidly to the top or bottom of the list can (though Firefox is easier to trigger the issue).

### Expected behavior

`isScrolling` should revert to `false` when scrolling has finished.

### How often does this bug happen?

Often

### Screenshots or Videos

[Kapture 2024-08-17 at 21.37.00.webm](https://github.com/user-attachments/assets/88ee70e5-a057-4510-8702-fbe377be275e)

### Platform

macOS 14.5
both Firefox (129.0.1) and Chrome (127.0.6533.119)

### tanstack-virtual version

3.8.6

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

PR #875, opened 2024-11-06
fix: add `useScrollendEvent` option

hshoja:
fixes https://github.com/TanStack/virtual/issues/799

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -133,7 +133,6 @@ const supportsScrollend =
 export const observeElementOffset = <T extends Element>(
   instance: Virtualizer<T, any>,
   cb: (offset: number, isScrolling: boolean) => void,
-  debounceFallbackEnabled = false,
 ) => {
   const element = instance.scrollElement
   if (!element) {
@@ -146,7 +145,7 @@ export const observeElementOffset = <T extends Element>(
 
   let offset = 0
   const fallback =
-    !debounceFallbackEnabled && supportsScrollend
+    instance.options.useScrollendEvent && supportsScrollend
       ? () => undefined
       : debounce(
           targetWindow,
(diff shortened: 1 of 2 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -133,6 +133,7 @@ const supportsScrollend =
 export const observeElementOffset = <T extends Element>(
   instance: Virtualizer<T, any>,
   cb: (offset: number, isScrolling: boolean) => void,
+  debounceFallbackEnabled = false,
```
piecyk, line 136: Don't need to pass it here, we can read it from `instance.options.scrollendEventEnabled`

packages/virtual-core/src/index.ts
```diff
@@ -334,6 +337,7 @@ export class Virtualizer<
   scrollElement: TScrollElement | null = null
   targetWindow: (Window & typeof globalThis) | null = null
   isScrolling = false
+  debounceFallbackEnabled = false
```
piecyk, line 340: I think `debounceFallbackEnabled` is to generic name, what about?
```suggestion
  scrollendEventEnabled = true
```

piecyk (commented):
Looks good, can you also update dosc regarding this new option.

packages/virtual-core/src/index.ts
```diff
@@ -334,6 +335,7 @@ export class Virtualizer<
   scrollElement: TScrollElement | null = null
   targetWindow: (Window & typeof globalThis) | null = null
   isScrolling = false
+  useScrollendEvent = true
```
piecyk, line 338: I think we should move it to options, then we need to update docs

packages/virtual-core/src/index.ts
```diff
@@ -144,15 +144,16 @@ export const observeElementOffset = <T extends Element>(
...
-    ? () => undefined
-    : debounce(
-        targetWindow,
-        () => {
-          cb(offset, false)
-        },
-        instance.options.isScrollingResetDelay,
-      )
+  const fallback =
+    instance.options.useScrollendEvent && supportsScrollend
```
piecyk, line 148: Let's also use the same logic in observeWindowOffset https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts#L192

piecyk (changes requested):

piecyk (approved):

piecyk:
Thanks @hshoja 🙌

PR #887, opened 2024-11-25
examples(vue): fix table example

tuanalumi:
This PR applies the same calculation from React to Vue example, which would fix:
- Lagging issue when scrolling
- Wrong row height


kouts (approved):
Thanks for the fix @tuanalumi

examples/vue/table/src/App.vue
```diff
@@ -47,9 +47,11 @@
           </thead>
           <tbody>
             <tr
-              v-for="virtualRow in virtualRows"
+              v-for="(virtualRow, index) in virtualRows"
               :key="virtualRow.key"
-              :style="{ transform: `translateY(${virtualRow.start}px)` }"
+              :style="{
+                transform: `translateY(${virtualRow.start - index * virtualRow.size}px)`,
```
piecyk, line 53: hmm this looks wrong, start should be calculated correctly

piecyk (changes requested):

examples/vue/table/src/App.vue
tuanalumi: I am not sure about what you're suggesting.

The calculation looks correct and that's the same with React version. I tested with this: https://stackblitz.com/github/tuanalumi/virtual/tree/main/examples/vue/table?embed=1&theme=light&preset=node&file=src/App.vue

PR #888, opened 2024-11-28
Updates faker method

schalkventer:
`.name` is no longer supported


piecyk (changes requested):
Thanks, @schalkventer can you update it also in other examples

Issue #573, opened 2023-08-14
miss item when set lanes to 4 and count is 36?


oloshe:
### Describe the bug

when lanes and count are set to some specific numbers, it may miss some items when displayed.
such as lanes = 4, count = 36
or lanes = 6, count = 36.

### Your minimal, reproducible example

https://codesandbox.io/s/hopeful-mcclintock-87sct9?file=/src/App.js:132-134

### Steps to reproduce

1. modify file `examples/react/src/main.tsx` line 8:
```typescript
const rows = new Array(36)
```
2. run `yarn start`
3. see the example `variable/Masonry (vertical)`

### Expected behavior

show all items

### How often does this bug happen?

None

### Platform

Windows 11

### tanstack-virtual version

3.0.0-beta.54

### TypeScript version

5.1.6

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

oloshe:
It's ok when set overscan equal to 5

DominikSerafin:
Might be related: https://github.com/TanStack/virtual/issues/665#issuecomment-2513499572

PR #933, opened 2025-02-22
fix: align startIndex & endIndex to lanes for masonry mode

Azq2:
1. startIndex should be aligned to the beginning of its lane
2. endIndex should be aligned to the end of its lane
3. count is actually lastIndex

Fix for #573

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -702,14 +702,15 @@ export class Virtualizer<
   )
 
   calculateRange = memo(
-    () => [this.getMeasurements(), this.getSize(), this.getScrollOffset()],
-    (measurements, outerSize, scrollOffset) => {
+    () => [this.getMeasurements(), this.getSize(), this.getScrollOffset(), this.options.lanes],
+    (measurements, outerSize, scrollOffset, lanes) => {
       return (this.range =
         measurements.length > 0 && outerSize > 0
           ? calculateRange({
               measurements,
               outerSize,
               scrollOffset,
+              lanes
             })
           : null)
     },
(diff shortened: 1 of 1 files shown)


Azq2:
Without fix
```
item.lane=
    | 1 | <-- row 0
| 0 | 1 | <-- row 1
| 0 | 1 | <-- row 2
```

With fix
```
item.lane=
| 0 | 1 | <-- row 0
| 0 | 1 | <-- row 1
| 0 | 1 | <-- row 2
```

Azq2:
Easy way for reproducing bug - just set overscan 0 for MasonryVerticalVirtualizerVariable example

https://tanstack.com/virtual/latest/docs/framework/react/examples/variable

packages/virtual-core/src/index.ts
```diff
@@ -1105,23 +1106,32 @@ function calculateRange({
...
 
   while (
-    endIndex < count &&
+    endIndex < lastIndex &&
     measurements[endIndex]!.end < scrollOffset + outerSize
   ) {
     endIndex++
   }
 
+  if (lanes != 1) {
```
piecyk, line 1129:
```suggestion
  if (lanes > 1) {
```

piecyk (commented):
Thanks, looks good, let's update the condition and we can merge it.

packages/virtual-core/src/index.ts
Azq2: Thanks for the feedback, I fixed it

piecyk:
@Azq2 thanks, btw did you noticed the formatting check?

Azq2:
@piecyk Yes, I missed it. Now it's fixed.

PR #864, opened 2024-10-27
fix: update scrollToIndex center-alignment

christriants:
- Fixes https://github.com/TanStack/virtual/issues/559

With the assumption that the current-behavior is not intended, I opened this PR to correct the `align: center` behavior of `scrollToIndex` to align the scroll item's center, not its top, with the center of the list.

Video of issue: 
https://github.com/user-attachments/assets/7b2445d7-e08e-4eee-aa97-3e454409214c

Video of fix: 
https://github.com/user-attachments/assets/607a619e-fc24-4fb4-8c9a-18ec35c917af

PR #935, opened 2025-02-24
fix(virtual-core): fixes center alignment option for scrollToOffset

### LINKED ISSUE #931: scrollToOffset with `align: 'center'` doesn't align correctly anymore
### Describe the bug

This is a regression introduced by PR #864, released in version 3.11.2. That change aimed to fix alignment when using `scrollToIndex`, but it has broken center alignment when using `scrollToOffset`.

When I call `scrollToOffset` with `align: center`, I expect it to anchor the scroll element so that the pixel offset is in the center. However, it now behaves the same as `align: start`.

https://stackblitz.com/edit/github-hsqwwxtq?file=src%2Fmain.tsx

### Steps to reproduce

1. Open the StackBlitz example above.
2. In the preview, click the "Scroll to Random Offset" button. This should scroll the list to the offset shown under the button.
3. BUG: The list container is aligned to the top of the offset, not the center, as illustrated by the blue marker line.

### Expected behavior

When `scrollToOffset(offset, { align: center })` is called, I expect the list container to be scrolled to where the offset position is in the center.

In the repro example, the blue marker line should be centered in the list container.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS 15.3.1
- Browser: Firefox 135.0.1, Chrome 133.0.6943.127

### tanstack-virtual version

3.13.0

### TypeScript version

5.7.2

### Additional context

If I downgrade to version 3.11.1 or earlier, the center alignment works correctly.

### Terms & Code of Conduct

soren121:
Fixes #931

This fixes a regression caused by PR #864, which adjusted the center alignment of `scrollToIndex` to center on the item. That change moved the center alignment adjustment from `getOffsetForAlignment` into `getOffsetForIndex`. However, `scrollToOffset` also relied on the former function, and now it lacks an implementation for `align: center`.

This PR restores the center alignment to `getOffsetForAlignment`, but adds a new `itemSize` parameter. This lets us share the center alignment implementation with both scroll functions, while also taking the item size into account when calculating the center alignment for `scrollToIndex`.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -873,17 +873,23 @@ export class Virtualizer<
     )
   }
 
-  getOffsetForAlignment = (toOffset: number, align: ScrollAlignment) => {
+  getOffsetForAlignment = (
+    toOffset: number,
+    align: ScrollAlignment,
+    itemSize = 0,
+  ) => {
     const size = this.getSize()
     const scrollOffset = this.getScrollOffset()
 
     if (align === 'auto') {
-      if (toOffset >= scrollOffset + size) {
-        align = 'end'
-      }
+      align = toOffset >= scrollOffset + size ? 'end' : 'start'
     }
 
-    if (align === 'end') {
+    if (align === 'center') {
+      // When aligning to a particular item (e.g. with scrollToIndex),
+      // adjust offset by the size of the item to center on the item
+      toOffset += (itemSize - size) / 2
+    } else if (align === 'end') {
       toOffset -= size
     }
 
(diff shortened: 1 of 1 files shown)


piecyk (approved):
Thanks, looks good 👍 When checking the getOffsetForAlignment current we calculate the max offset via reading scrollWidth/scrollHeight, kinda feels we can use the totalSize so just

```tsx
const maxOffset = this.getTotalSize() - size
```

and drop the scrollSize.

soren121:
> When checking the getOffsetForAlignment current we calculate the max offset via reading scrollWidth/scrollHeight, kinda feels we can use the totalSize

@piecyk Yeah, that seems like a reasonable improvement to me.

PR #937, opened 2025-02-27
fix(virtual-core): expand range in masonry layouts to catch items from all lanes

### LINKED ISSUE #665: Total size is calculated incorrectly for masonry grids
### Describe the bug

I'm using @tanstack/svelte-virtual. With masonry type grids total size (`.getTotalSize()`) seems to be based on all other row heights + the height of the shortest item in the last row of the grid (instead tallest item in the latest row of the grid).

### Expected behavior

For masonry type grids `.getTotalSize()` being calculated based on tallest item in each row instead of shortest item for the very last row.

### Platform

macOS, Chrome

### tanstack-virtual version

v3.0.4

### Terms & Code of Conduct

### LINKED ISSUE #936: Range is calculated incorrectly for masonry layouts, causing items to be missed
### Describe the bug

I think this is likely another issue related to #573.

While debugging an issue plaguing my project, I've noticed that `calculateRange` does not properly account for masonry layouts. It assumes that visible items will be directly adjacent in the `measurements` array, and while this is true for linear layouts, it's not true for masonry layouts.

For example, imagine a masonry layout with 3 lanes (this is the same as in the attached reproduction.) What if the first lane has an item that is much longer than the items in the other two lanes? Let's examine what will happen:
1. `calculateRange` will find the item in `measurements` that has a start offset closest to the scroll offset.
2. Then, it will continue incrementing indices until it finds an item whose end offset is past the virtualizer window.

The problem is, items from different lanes are interleaved in `measurements`. If I scroll past the start of the long item, it's very likely that items in the other lanes will have a start offset that's closer to the scroll offset. If that happens, the long item probably won't be included in the returned range, and so it will disappear too early.

There are a couple approaches I can think of to solve this:
1. We could calculate ranges _per lane_, and return an array of ranges. This would be a breaking change, because we expose `range` and `rangeExtractor` in the public API.
2. We maintain a running list of visible items, and add/evict items as needed when the scroll offset changes. Compared to the current approach where the range is always calculated "from scratch", this would avoid removing items too early, but I think it could be more challenging to implement. This would probably be a breaking change as well.

https://stackblitz.com/edit/tanstack-virtual-wg3vpdoo

### Steps to reproduce

1. Open the StackBlitz reproduction.
2. Scroll down in the preview pane.

Items will disappear from view before they are fully outside the viewport. Very large items are especially noticeable-- they will disappear from view much too early.

### Expected behavior

Masonry items should be visible until they are fully outside the viewport.

Every time. It's most obvious when the items vary greatly in size, and less obvious if the items are all similar sizes.

### Screenshots or Videos

The background color of the example is black. Areas that appear black are gaps where masonry items should be visible, but aren't.

https://github.com/user-attachments/assets/8ee7319b-8f97-43bb-b999-495264c631c9

### Platform

- OS: macOS 15.3.1
- Browser: Firefox 135.0.1

### tanstack-virtual version

3.13.2

### TypeScript version

5.7.2

### Additional context

This bug is most apparent when `overscan` is set to 0.

One way to work around it is to increase `overscan`, but this doesn't actually alleviate the problem, and depending on how large your items are, you may still have "missed" items.

### Terms & Code of Conduct

soren121:
Fixes #936 
Fixes #665 

This PR resolves two issues related to items being "missed" (i.e. unexpectedly cut-off or not rendered) in masonry layouts.

The first problem is that `calculateRange` was not changed when we added support for masonry layouts. This function assumes that items are purely sequential in the layout: it finds _an_ item close to the scroll offset, and expands forward until it finds an item that exceeds the window. Masonry layouts do not behave like this. Even though our `measurements` array is one-dimensional, the view is two-dimensional. Items which are far apart in `measurements` may still be rendered side-by-side, and that needs to be accounted for.

The solution I came up with is to expand the range in both directions until we've found every item that should be in-view in each lane. This feels somewhat naive, and perhaps there's a more performant solution, but I think for the majority of masonry layouts, it should perform just fine.

The second problem is in the same vein: `getTotalSize` misses some items because it's _not_  always true that the last items in each lane will be the last N items in `measurements` (where N is the number of lanes.) This is because `measurements` is sorted by the `start` position, and if you have very uneven item sizes, the `end` positions could be very different from each other. I changed this to work in a similar way: it scans from the bottom up until it's found end positions for all lanes, and then selects the maximum.

Both of the issues can be seen in the reproduction link in #936.


soren121:
I closed this temporarily because I was still hitting problems in my application, but I was able to verify that the bug was just in my application and not this PR.

piecyk (approved):
Looks great, thanks @soren121 🥇

PR #948, opened 2025-03-09
docs: add documentation for sync parameter of the onChange function.

taylous:
## summary

the description for the sync parameter of the `onChange` function was missing from the API documentation,
so I've added it.

upon reviewing the internal code, i found that it is invoked by the `notify` function.
> specifically through: `maybeNotify`, `resizeItem`, `measure`

## references

https://github.com/TanStack/virtual/blob/721bf538386482d201c1e1bff1a976c5997a90c9/packages/virtual-core/src/index.ts#L447

https://github.com/TanStack/virtual/blob/721bf538386482d201c1e1bff1a976c5997a90c9/packages/virtual-core/src/index.ts#L462

https://github.com/TanStack/virtual/blob/721bf538386482d201c1e1bff1a976c5997a90c9/packages/virtual-core/src/index.ts#L827

https://github.com/TanStack/virtual/blob/721bf538386482d201c1e1bff1a976c5997a90c9/packages/virtual-core/src/index.ts#L1072

### DIFF
--- a/docs/api/virtualizer.md
+++ b/docs/api/virtualizer.md
@@ -67,7 +67,7 @@ The initial `Rect` of the scrollElement. This is mostly useful if you need to ru
 ### `onChange`
 
 ```tsx
-onChange?: (instance: Virtualizer<TScrollElement, TItemElement>) => void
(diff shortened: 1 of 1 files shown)


piecyk:
Thanks

PR #955, opened 2025-03-13
fix(virtual-core): ensure all items render when content height is less than or equal to viewport height

### LINKED ISSUE #871: react-virtual – Visible items are incorrect on first render when using `max-height` on container
### Describe the bug

When using a dynamically sized list with a container with `max-height`, the first render of the virtualizer correctly sets the height of the list after measuring the elements, but does not show the correct number of items. If another render is forced, then the virtualizer updates with the correct number of items. See the attached video and https://codesandbox.io/p/sandbox/tanstack-virtual-react-max-height-bug-forked-wtscyh?workspaceId=d07a637a-3808-4246-a35e-982f2bc7080c

https://codesandbox.io/p/sandbox/tanstack-virtual-react-max-height-bug-forked-wtscyh?workspaceId=d07a637a-3808-4246-a35e-982f2bc7080c

### Steps to reproduce

1. Open https://codesandbox.io/p/sandbox/tanstack-virtual-react-max-height-bug-forked-wtscyh?workspaceId=d07a637a-3808-4246-a35e-982f2bc7080c
2. See that the virtualizer is showing only 2 items, when more can fit
3. Wait 3s, then observe when re-render is forced that virtualizer now shows the correct number of items

### Expected behavior

When the virtualizer is used with dynamically sized elements and a container with `max-height` instead of a fixed height, after measure some elements it should show the correct number of visible items needed to fill the container space.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/f4528744-eb03-41ae-91ef-ce8923a2a264

### Platform

OSX Chrome

### tanstack-virtual version

v3.10.8

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

SaeWooKKang:
Fixes #871 

## Problem:
This issue occurs because when the scroll container size (scrollRect) changes and the `maybeNotify` function is called, scrollRect is not included in the dependency array, causing the virtual item range to not be calculated correctly.

## Solution:
I added scrollRect value to the dependency array of the `maybeNotify` function, ensuring that the virtual item range is recalculated correctly when the scroll container size changes.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -456,7 +456,7 @@ export class Virtualizer<
         this.isScrolling,
         this.range ? this.range.startIndex : null,
         this.range ? this.range.endIndex : null,
-        this.scrollRect 
+        this.scrollRect,
       ]
     },
     (isScrolling) => {
@@ -469,7 +469,7 @@ export class Virtualizer<
         this.isScrolling,
         this.range ? this.range.startIndex : null,
         this.range ? this.range.endIndex : null,
-        this.scrollRect 
+        this.scrollRect,
       ] as [boolean, number | null, number | null, Rect],
     },
   )
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -456,6 +456,7 @@ export class Virtualizer<
         this.isScrolling,
         this.range ? this.range.startIndex : null,
         this.range ? this.range.endIndex : null,
+        this.scrollRect 
       ]
     },
     (isScrolling) => {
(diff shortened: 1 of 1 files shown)


SaeWooKKang:
umm.. I think the test fails because the number of renders increases by one in addition to the dependency array. 

Is there a better way to fix the bug without fixing the test?

packages/virtual-core/src/index.ts
```diff
@@ -456,6 +456,7 @@ export class Virtualizer<
         this.isScrolling,
         this.range ? this.range.startIndex : null,
         this.range ? this.range.endIndex : null,
+        this.scrollRect,
```
piecyk, line 459: hmm we are reading the scrollRect in calculateRange via getSize, so if the range should be correct

piecyk:
Thanks, @SaeWooKKang, for your work on this! I’ve found a better approach to address this issue in #957, so I’ll be closing this PR in favor of that solution. I really appreciate your effort!.

SaeWooKKang:
@piecyk I'm glad it was solved in a better way. Thank you for your efforts!

PR #964, opened 2025-03-24
fix(core): handle case when item count is less than or equal to lanes

### LINKED ISSUE #963: getTotalSize() returns 0 when using multiple lanes with a single item
### Describe the bug

Starting from version `3.13.3`, the behavior of `virtualizer.getTotalSize()` has changed. When using multiple `lanes` (i.e., `lanes > 1`) and the item `count` is `1`, `getTotalSize()` now returns `0`.
This did not happen in `3.13.2`, where the total size of a single item was calculated and returned correctly.

If this change in behavior is intentional, please let me know.

https://stackblitz.com/edit/tanstack-virtual-4zlr1xau

### Steps to reproduce

1. start app `npm install && npm run start`
2. variable settings
    - `@tanstack/react-virtual` version: 3.13.2
    - lanes: 2
    - contents count: 1
3. change `VERSION`(`main.tsx#L8`) to `"3.13.3"` and observe that the same setup returns the expected item size.

### Expected behavior

with `count === 1` and `lanes > 1`, `getTotalSize()` should return the size of the single item, not `0`.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/633c9468-f255-4466-b025-c975363079fb

### Platform

- OS: macOS 13.3.1 22E261 arm64
- Browser: Chrome 134.0.6998.89 (arm64)

### tanstack-virtual version

v3.13.3

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

Ryanromaris:
### Summary

This ensures calculateRange returns a valid range when there are fewer items than lanes (including single-item virtual lists). Prevents invalid access to undefined measurements and avoids empty or broken virtual ranges.

Fixes #963

### DIFF
--- a/.changeset/tall-masks-call.md
+++ b/.changeset/tall-masks-call.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/virtual-core": patch
+'@tanstack/virtual-core': patch
 ---
 
 fix(core): handle case when item count is less than or equal to lanes
--- a/.changeset/tall-masks-call.md
+++ b/.changeset/tall-masks-call.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/virtual-core": patch
+---
+
+fix(core): handle case when item count is less than or equal to lanes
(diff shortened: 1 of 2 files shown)


piecyk:
Thanks @Ryanromaris 👏

PR #967, opened 2025-03-28
fix(virtual-core): fix total size calculation for single item in multi-lane

### LINKED ISSUE #963: getTotalSize() returns 0 when using multiple lanes with a single item

greenblues1190:
This is a follow-up to #961. The original issue still occurs when there's only one item in a multi-lane layout.

This PR changes the loop condition in `getTotalSize()` to handle cases where only one item exists in multi-lane. The previous condition skipped the loop when `endIndex === 0`, which caused the total size to be incorrectly calculated as 0.

I used the following test to verify the fix locally:

```ts
test("should return correct total size with single item in multi-lane", () => {
  const virtualizer = new Virtualizer({
    count: 1,
    lanes: 2,
    getScrollElement: () => null,
    estimateSize: () => 50,
    scrollToFn: vi.fn(),
    observeElementRect: vi.fn(),
    observeElementOffset: vi.fn(),
  });
  expect(virtualizer.getTotalSize()).toBe(50);
});
```

This is a fairly specific edge case, so I didn’t include the test in the commit.
Happy to add it if you think it’d be useful to include in the suite!

Fixes #963

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -1162,7 +1162,7 @@ function calculateRange({
     // Expand backward until we include all lanes' visible items
     // closer to the top
     const startPerLane = Array(lanes).fill(scrollOffset + outerSize)
-    while (startIndex > 0 && startPerLane.some((pos) => pos >= scrollOffset)) {
+    while (startIndex >= 0 && startPerLane.some((pos) => pos >= scrollOffset)) {
       const item = measurements[startIndex]!
       startPerLane[item.lane] = item.start
       startIndex--
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -1046,7 +1046,7 @@ export class Virtualizer<
     } else {
       const endByLane = Array<number | null>(this.options.lanes).fill(null)
       let endIndex = measurements.length - 1
-      while (endIndex > 0 && endByLane.some((val) => val === null)) {
+      while (endIndex >= 0 && endByLane.some((val) => val === null)) {
         const item = measurements[endIndex]!
         if (endByLane[item.lane] === null) {
           endByLane[item.lane] = item.end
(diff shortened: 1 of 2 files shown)


piecyk:
Thanks, @greenblues1190! Let's add the test. Overall, the tests need some attention. The plan is to refactor the current unit tests a bit and add browser testing using Playwright.

soren121 (commented):
If this is an off-by-one, then the condition on L1165 also needs to be adjusted (it's essentially the same loop.)

greenblues1190:
@piecyk Test added!

@soren121 You're right, I’ve updated the condition too.

PR #986, opened 2025-04-25
fix(virtual-core): prevent measurement jitter when scale is applied

usings:
### Problem
When using the `measureElement` function in combination with scaled elements (via CSS `transform: scale(...)`), the `getBoundingClientRect()` method returns fractional values that may cause layout jitter or re-renders due to rounding inconsistencies.

### Solution
Fallback measurement now uses `offsetWidth` / `offsetHeight` instead of `getBoundingClientRect()` when `ResizeObserverEntry.borderBoxSize` is not available. This avoids jitter caused by scaled bounding boxes.

### Why `offset*`?
- `offsetWidth` / `offsetHeight` give integer values.
- Not affected by `transform: scale(...)`, unlike `getBoundingClientRect()`.
- More stable for virtualizer calculations when dealing with scaled elements.

### Impact
Improves measurement stability in virtualized lists using CSS transforms. No breaking changes.


piecyk:
@usings Could you also update observeElementRect to use offsetWidth / offsetHeight as a fallback instead of getBoundingClientRect()?

usings:
> @usings Could you also update observeElementRect to use offsetWidth / offsetHeight as a fallback instead of getBoundingClientRect()?

Sure. I’ll make that change shortly.

packages/virtual-core/src/index.ts
```diff
@@ -74,10 +79,10 @@ export const observeElementRect = <T extends Element>(
 
   const handler = (rect: Rect) => {
     const { width, height } = rect
-    cb({ width: Math.round(width), height: Math.round(height) })
```
piecyk, line 77: Let's round the values for borderBoxSize, which can still return fractional values.

piecyk (changes requested):

packages/virtual-core/src/index.ts
usings: Got it!

piecyk:
Thanks @usings

PR #993, opened 2025-05-05
fix(react): table sorting example delay

giuliobracci:
This commit ensures a fluid user experience when applying sorting on TanStack table by reading the deferred value instead of the actual state.

### DIFF
--- a/examples/react/table/src/main.tsx
+++ b/examples/react/table/src/main.tsx
@@ -9,12 +9,13 @@ import {
   useReactTable,
 } from '@tanstack/react-table'
 import { makeData } from './makeData'
-import type { ColumnDef, Row, SortingState } from '@tanstack/react-table'
+import type { ColumnDef, SortingState } from '@tanstack/react-table'
 import type { Person } from './makeData'
 import './index.css'
 
 function ReactTableVirtualized() {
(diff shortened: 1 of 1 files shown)


examples/react/table/src/main.tsx
```diff
@@ -9,12 +9,13 @@ import {
...
 import { makeData } from './makeData'
-import type { ColumnDef, Row, SortingState } from '@tanstack/react-table'
+import type { ColumnDef, SortingState } from '@tanstack/react-table'
 import type { Person } from './makeData'
 import './index.css'
 
 function ReactTableVirtualized() {
-  const [sorting, setSorting] = React.useState<SortingState>([])
+  const [_sorting, setSorting] = React.useState<SortingState>([])
+  const sorting = React.useDeferredValue(_sorting);
```
piecyk, line 18: @giuliobracci Hmm, virtualized tables already improve performance by only rendering visible rows. I'm a bit concerned this might give the impression that useDeferredValue is needed in most cases, which isn't really true.

giuliobracci: Hi @piecyk you can try the diff with the example. 
To me through [stackblitz](https://tanstack.com/virtual/latest/docs/framework/react/examples/table) it seems that the lag is very noticeable when applying sorting.

PR #995, opened 2025-05-05
fix(virtual-core): loosen approxEqual to allow 1px difference

### LINKED ISSUE #980: scrollToIndex blocks scrolling after execution in Safari
### Describe the bug

I use useWindowVirtualizer hook for making virtual elements and scrollToIndex with dynamically-sized elements and { align: "start" }.
After scrolling the page manually (to allow the virtualizer to cache element heights), clicking the “Scroll to element” button scrolls correctly to the element’s header. However, after that, scrolling becomes impossible — the page keeps snapping back to the element’s header, effectively locking the scroll.

Note: I found a similar issue https://github.com/TanStack/virtual/issues/473

https://codesandbox.io/p/devbox/heuristic-rgb-qwt484?file=%2Fsrc%2Fmain.tsx%3A34%2C16

### Steps to reproduce

Zoom the page in Safari (e.g., 120%).
Scroll to an element with index 70.
Press the "scroll to element" button.
Try to scroll normally.

### Expected behavior

After scrolling to the element, the user should be able to scroll the page freely.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/13952f50-4460-4fa8-9101-e8ce013a7cf1

### Platform

I reproduced it in
Safari - 15.6.1, 18.2, 18.3.1

### tanstack-virtual version

3.13.6

### TypeScript version

_No response_

### Additional context

This issue seems to happen when the page is zoomed in Safari. Safari can render line-wrapped text elements with fractional heights, which may cause inconsistencies in the virtualizer’s height calculations and result in incorrect scroll positioning or locking.

[This](https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts#L1010) code is triggered in an infinite loop because accumulated height inaccuracies cause a mismatch with the initialOffset (window.scrollY) value. I'd expected to break this loop if it made more than N calls with the same values.

### Terms & Code of Conduct

piecyk:
fixes #980

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -904,16 +904,7 @@ export class Virtualizer<
       toOffset -= size
     }
 
-    const scrollSizeProp = this.options.horizontal
-      ? 'scrollWidth'
-      : 'scrollHeight'
-    const scrollSize = this.scrollElement
-      ? 'document' in this.scrollElement
-        ? this.scrollElement.document.documentElement[scrollSizeProp]
-        : this.scrollElement[scrollSizeProp]
-      : 0
-
-    const maxOffset = scrollSize - size
+    const maxOffset = this.getTotalSize() - size
 
     return Math.max(Math.min(maxOffset, toOffset), 0)
   }
@@ -1010,8 +1001,8 @@ export class Virtualizer<
           const [latestOffset] = notUndefined(
             this.getOffsetForIndex(index, align),
           )
-
-          if (!approxEqual(latestOffset, this.getScrollOffset())) {
+          const currentScrollOffset = this.getScrollOffset()
+          if (!approxEqual(latestOffset, currentScrollOffset)) {
             this.scrollToIndex(index, { align, behavior })
           }
         } else {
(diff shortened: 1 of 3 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -1010,8 +1001,8 @@ export class Virtualizer<
           const [latestOffset] = notUndefined(
             this.getOffsetForIndex(index, align),
           )
-
-          if (!approxEqual(latestOffset, this.getScrollOffset())) {
+          const currentScrollOffset = this.getScrollOffset()
+          if (!approxEqual(latestOffset, currentScrollOffset)) {
```
ogonkov, line 1005: Looks like there still a chance to get infinite loop

piecyk: Do you have something specific in mind?

ogonkov: I guess for any given `scrollToIndex` we want to set some kind of counter, and break recursion when this counter is reach some limit, kinda dead loop detector

Issue #910, opened 2025-01-07
Note that block jumps in unintuitive ways.


kmpeeduwee:
### Describe the bug

When the top face of the block is at the top (invisible), pressing the button adds a blue rectangle, but the block itself grows upward while its lower part remains in place. It should be the opposite. If the block is fully visible, everything works as it should.

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/gallant-sun-f6gmp9?workspaceId=ws_VgNsJVVZEmm7CqrvpZ8GN2

### Steps to reproduce

1 Move the block that we want to expand so that its top part is not visible.
2 Press the Expand button.

### Expected behavior

The block should expand downward.

### How often does this bug happen?

Every time

### Screenshots or Videos

Normal behavior (full block visible) :
https://github.com/user-attachments/assets/2b75cf6e-de70-460e-b476-6312516511f6

Non-normal behavior (part block visible) :
https://github.com/user-attachments/assets/9b2bf3af-2d96-404e-ba17-bdc89d1447bc

### Platform

Linux Fedora 41, mac os 15, firefox 133, chrome 131

### tanstack-virtual version

3.10.9, 3.11.2

### TypeScript version

5.7.2

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

kmpeeduwee:
@Azq2 Thank you! Unfortunately, it's not work. I fixed example https://codesandbox.io/p/devbox/gallant-sun-f6gmp9?workspaceId=ws_VgNsJVVZEmm7CqrvpZ8GN2

piecyk:
@kmpeeduwee it's default behaviour, can be disabled using shouldAdjustScrollPositionOnItemSizeChange, please checkout docs https://tanstack.com/virtual/latest/docs/api/virtualizer#shouldadjustscrollpositiononitemsizechange

Overall we should change the logic to adjust position only on scrolling backward.

Issue #562, opened 2023-07-11
Expanding an item that is partially obscured by the top of the scrolling window causes the item to move upwards


adlai-arnold:
### Describe the bug

Expanding an item should never move the top of the item--it should always expand downwards. However, if the scroll window occludes the top of the item, it will instead move upwards when expanded. 

**This issue does not occur in `3.0.0-beta.35` and earlier.**

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/zen-blackburn-tnrzfr

### Steps to reproduce

1. Scroll down slightly, so that the first item is partially occluded by the top of the scroll window
2. Click the first item to expand it
3. Note that the top of the item jumps up and out of view

### Expected behavior

As a user, I expect that expanding an item will not cause it to move.

### How often does this bug happen?

Every time

### Platform

MacOS 13.4.1 (22F82), Chrome Version 114.0.5735.198 (Official Build) (arm64)

### tanstack-virtual version

3.0.0-beta.54

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

adlai-arnold:
Given that the issue starts in beta release 36, it must be somewhere in this diff: https://github.com/TanStack/virtual/compare/v3.0.0-beta.35...v3.0.0-beta.36

reidark:
Any solutions/workaround on that? 

I have a virtual list of checkboxes and every time the user select some item that is not 100% on the "viewport", it scrolls upwards, causing a confusion.

This happens because of the focus state? I know if I navigate via keyboard the list will scroll automatically based on the element focused.

piecyk:
No it is happening because of our auto adjustment scroll position logic, basic if user jumps to bottom of the list an start scrolling up, in dynamic case the sizes can change, that will change the height of the list making the scroll to jump


https://github.com/TanStack/virtual/blob/beta/packages/virtual-core/src/index.ts#L656-L665

with this we are trying to limit that behaviour, best would to be to add it only when scrolling backward and maybe to allow to opt out from it.

kamuiiiii:
I've encountered the same issue. Any updates on this?

piecyk:
> I've encountered the same issue. Any updates on this?

Yeah, now it's possible to control the adjustment with `shouldAdjustScrollPositionOnItemSizeChange`, one option can be 

```jsx
[...]

PR #1002, opened 2025-05-14
Adapt default logic to adjust scroll position item size change

balzdur:
# Context

This change is based on those issues: #910, #562 
It tries to implement the idea from this message:
> @kmpeeduwee it's default behaviour, can be disabled using shouldAdjustScrollPositionOnItemSizeChange, please checkout docs https://tanstack.com/virtual/latest/docs/api/virtualizer#shouldadjustscrollpositiononitemsizechange
> 
> Overall we should change the logic to adjust position only on scrolling backward. 

 _Originally posted by @piecyk in [#910](https://github.com/TanStack/virtual/issues/910#issuecomment-2679300004)_
 
Disclaimer:  I'm not an expert of the lib, neither someone used to contribute a lot to open source (and/or on tanstack repositories)

### DIFF
--- a/.changeset/petite-carrots-beam.md
+++ b/.changeset/petite-carrots-beam.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/virtual-core': minor
+---
+
+feat(virtual-core): change default condition to adjust scroll position on item size change
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -814,7 +814,7 @@ export class Virtualizer<
       if (
         this.shouldAdjustScrollPositionOnItemSizeChange !== undefined
           ? this.shouldAdjustScrollPositionOnItemSizeChange(item, delta, this)
-          : item.start < this.getScrollOffset() + this.scrollAdjustments
+          : this.scrollDirection === "backward" && item.start < this.getScrollOffset() + this.scrollAdjustments
       ) {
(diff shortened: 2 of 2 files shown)


balzdur:
Is it ok to merge this or do I need to make some stuff ? 🙂

piecyk:
@balzdur looks like with this change we broke bidirectional lists, for example, in chat apps where scrolling upward loads older messages. Even without actively scrolling, we should stay in the same place after new elements are loaded. See this related discussion: https://github.com/TanStack/virtual/discussions/1013#discussioncomment-13411874

diogowbrito:
@piecyk heads up that after updating Tanstack Virtual minor version (which includes this change), we noticed a different scrolling behavior in our application. We have a scenario in which we scroll to the bottom of the table and that stopped working properly (the list was no longer scrolling all the way to the bottom, leaving a few rows left to display. Our list loads data asynchronously (I'd say similarly to the bidirectional lists you mentioned in last comment).

This should have likely been included as a minor version update, instead of a bug fix, due to the behavior change.

Thanks for all the work!

piecyk:
@diogowbrito Hey, sorry for the hassle this caused, I appreciate you pointing it out. You're right, this change introduced a behavioral shift that probably should’ve gone out as a minor version instead. I’ll go ahead and revert it to avoid breaking existing use cases.

PR #1069, opened 2025-10-22
Fixed incorrect table height calculation, fixed some typing errors, reduced the amount of generated data to 1000

ArthurKoba:
## 🎯 Changes

When I redesigned my table based on the example, I noticed that the final height of the table was incorrectly determined. I started looking into it and it turned out that the parent link was set to the wrong element. When I installed it on a table element, the calculations began to proceed correctly. It's strange that during the last fix (https://github.com/TanStack/virtual/pull/887) they didn't find this problem and added a crutch :D

I also corrected some linter errors, but I don’t have time to deal with the typing of the sorting function, we’ll leave the decision to the next contributor :)

And I reduced the amount of generated data to 1000, I think it is obvious that there will be no difference between 1000 and 50000 in the display, but the page will take much longer to load.

https://github.com/TanStack/virtual/pull/887

## ✅ Checklist

### DIFF
--- a/examples/vue/table/src/App.vue
+++ b/examples/vue/table/src/App.vue
@@ -6,66 +6,61 @@
       the translateY pixel count differently and base it off the index.
     </p>
 
-    <div ref="parentRef" class="container">
-      <div :style="{ height: `${totalSize}px` }">
-        <table>
-          <thead>
-            <tr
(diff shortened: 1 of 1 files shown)


ArthurKoba:
Dear friends, I apologize, but it turns out my changes disabled virtualization completely.

PR #1080, opened 2025-11-22
fix: stabilize lane assignments in masonry layout

seob717:
Added lane assignment caching to prevent items from jumping between lanes when viewport is resized. Previously, items could shift to different lanes during resize due to recalculating "shortest lane" with slightly different heights.

## 🎯 Changes
- Added `laneAssignments` cache (Map<index, lane>) to persist lane assignments
- Lane cache is cleared when `lanes` option changes or `measure()` is called
- Lane cache is cleaned up when `count` decreases (removes stale entries)
- Lane cache is cleared when virtualizer is disabled

## ✅ Checklist

### DIFF
--- a/.changeset/stable-lane-assignments.md
+++ b/.changeset/stable-lane-assignments.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/virtual-core": patch
+'@tanstack/virtual-core': patch
 ---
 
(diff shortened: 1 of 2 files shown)


piecyk:
@seob717 Looks great overall, the lane caching + settling logic makes the masonry layout much more stable on resize.

PR #1085, opened 2025-12-03
fix(virtual-core): Notify framework when count changes to update getTotalSize()

shunn2:
## 🎯 Changes

Fixes issue where getTotalSize() returns stale values when count changes during filtering/search, causing blank space or inaccessible items.

Added onChange callback to getMeasurementOptions memo that auto-notifies framework when count/measurement options change.

## ✅ Checklist

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -646,9 +646,10 @@ export class Virtualizer<
     },
     {
       key: false,
+      skipInitialOnChange: true,
       onChange: () => {
         // Notify when measurement options change as they affect total size
-        this.notify(false)
+        this.notify(this.isScrolling)
       },
     },
   )


packages/virtual-core/src/index.ts
```diff
@@ -646,6 +646,10 @@ export class Virtualizer<
     },
     {
       key: false,
+      onChange: () => {
+        // Notify when measurement options change as they affect total size
+        this.notify(false)
```
piecyk:
```suggestion
        this.notify(this.isScrolling)
```

packages/virtual-core/src/index.ts
```diff
@@ -646,6 +646,10 @@ export class Virtualizer<
...
+      },
```
piecyk: @shunn2 I created a PR earlier (https://github.com/TanStack/virtual/pull/1088) that adds an initialization guard to the memo system so we avoid the extra initial `notify` call. This keeps the `onChange` semantics consistent (only firing when state actually changes), preserves the existing React re-render count expectations, and still fixes the stale `getTotalSize` issue.

With that now merged into `master`, you can simplify this block to

```suggestion
      skipInitialOnChange: true
      onChange: () => {
        // Notify when measurement options change as they affect total size
        this.notify(this.isScrolling)
      },
```

piecyk (changes requested):
Thanks for the PR, great work!

Please rebase onto the latest `master` and update the memo to use `skipInitialOnChange` (added in #1088). This will ensure we avoid the extra initial `notify` while preserving correct `onChange` semantics.

After that, we can merge.

LynxyssCZ:
The docs should be updated, as the `getItemKey` now **needs** to be memoized in order to avoid infinite re-render loop in React, causing a crash.

The current version states `This function should be memoized to prevent unnecessary re-renders.`, which is not exactly true after this change.

piecyk:
@LynxyssCZ Good catch, you’re right. With the recent changes, getItemKey needs to be memoized to avoid an infinite render loop in React.

Thanks for pointing this out. We’ll get the documentation updated to reflect the new behavior more accurately. We’re also happy to accept a PR if you’d like to contribute.

piecyk:
@LynxyssCZ updated in https://github.com/TanStack/virtual/pull/1103

Issue #659, opened 2024-02-01
Scrolling up with dynamic heights stutters and jumps


nconfrey:
### Describe the bug

I have a feed of dynamic items, like iframes, photos, and text. When I scroll downward, everything works great and the scrolling is smooth, as measured items that increase in height push the other items down out of sight. However, when I scroll upwards, the performance is super stuttery and the items jump all over the place:

https://github.com/TanStack/virtual/assets/7350670/655b0a0b-4562-47e2-aa96-cb72daf8ad37

### Your minimal, reproducible example

Code below:

### Steps to reproduce

Create the virtualizer as normal:
```
const rowVirtualizer = useVirtualizer({
    count: itemCount,
    getScrollElement: () => parentRef.current,
    estimateSize: () => 100,
    overscan: 2
  })
```
Then the feed:
```
const mainFeed = () => {
    return (
      <div
        style={{
          height: `${rowVirtualizer.getTotalSize()}px`,
          width: '100%',
          position: 'relative',
        }}
      >
        {rowVirtualizer.getVirtualItems().map((virtualRow) => {
          const post = loadedPosts[virtualRow.index]
          return (
            <div
              key={virtualRow.key}
              data-index={virtualRow.index}
              ref={rowVirtualizer.measureElement}
              style={{
                position: 'absolute',
                top: 0,
                left: 0,
                width: '100%',
                transform: `translateY(${virtualRow.start - rowVirtualizer.options.scrollMargin}px)`,
                display: 'flex',
              }}
            >
              <div style={{ width: '100%' }}>
                <FeedItem post={post} />
              </div>
            </div>
          )
        })}
      </div>
    )
  }
```

### Expected behavior

Scrolling upwards should be as smooth as scrolling downwards.

### How often does this bug happen?

Every time

### Platform

macOS, Chrome

### tanstack-virtual version

"@tanstack/react-virtual": "^3.0.1"

### Additional context

The following code inside of the `useVirtualizer` fixes the issue:
```
    measureElement: (element, entry, instance) => {
      const direction = instance.scrollDirection
      if (direction === "forward" || direction === null) {
        return element.scrollHeight
      } else {
        // don't remeasure if we are scrolling up
        const indexKey = Number(element.getAttribute("data-index"))
        let cacheMeasurement = instance.itemSizeCache.get(indexKey)
        return cacheMeasurement
      }
    }
```

I propose that the above behavior is default, that items are not remeasured when scrolling upward.

### Terms & Code of Conduct

[...]

PR #1089, opened 2025-12-08
feat(virtual-core): add skipRemeasurementOnBackwardScroll option to reduce stuttering

dpwls02142:
## 🎯 Changes
1. problem
- During heavy scrolling with dynamic item sizes, frequent scroll adjustments can cause:
  - Layout thrashing due to multiple synchronous scroll position updates
  - Stuttering during backward scrolling because already-measured items are remeasured unnecessarily
- Related issue: #659

2. Solution
- Batching Scroll Adjustments
  - Collect multiple adjustment deltas in `pendingAdjustmentDeltas` array
  - Use `requestAnimationFrame` to batch-apply all adjustments in a single frame
  - Reduces layout thrashing and improves scrolling performance

- Backward Scroll Optimization
  - Add `disableScrollAdjustmentOnBackwardScroll` option
  - Skip remeasuring items that are already in cache during backward scroll
  - Prevents stuttering while maintaining measurement accuracy

## ✅ Checklist

## 🚀 Release Impact

- [x] This change is docs/CI/dev-only (no release).

### DIFF
--- a/.changeset/witty-bears-fail.md
+++ b/.changeset/witty-bears-fail.md
@@ -1,0 +1,14 @@
+---
+'@tanstack/virtual-core': minor
+---
+
+feat: add skipRemeasurementOnBackwardScroll option to reduce stuttering
+
+Adds new option to skip re-measuring already-cached items during backward scrolling. 
+This prevents scroll adjustments that conflict with the user's scroll direction, reducing stuttering in dynamic content like social feeds or chat messages.
+
+When enabled, cached measurements are reused during backward scroll while `isScrolling` is true. Layout settles correctly once scrolling stops.
+
+**Changes:**
+- Added `skipRemeasurementOnBackwardScroll` option (default: `false`)
+- Skip re-measurement in `_measureElement` when scrolling backward with cached items
--- a/docs/api/virtualizer.md
+++ b/docs/api/virtualizer.md
@@ -270,6 +270,15 @@ This option enables wrapping ResizeObserver measurements in requestAnimationFram
 
 It helps prevent the "ResizeObserver loop completed with undelivered notifications" error by ensuring that measurements align with the rendering cycle. This can improve performance and reduce UI jitter, especially when resizing elements dynamically. However, since ResizeObserver already runs asynchronously, adding requestAnimationFrame may introduce a slight delay in measurements, which could be noticeable in some cases. If resizing operations are lightweight and do not cause reflows, enabling this option may not provide significant benefits.
 
+### `skipRemeasurementOnBackwardScroll`
+
+```tsx
+skipRemeasurementOnBackwardScroll: boolean
+```
+When enabled, prevents re-measuring items that have already been measured during backward scrolling.
+This reduces stuttering caused by scroll position adjustments that conflict with the user's scroll direction.
+It is recommended to use this property when scrolling in situations where item heights change dynamically.
+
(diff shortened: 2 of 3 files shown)


piecyk:
@dpwls02142 Thanks for the PR, this is addressing a very real problem. I think there are actually two distinct changes here:

1. Skipping re-measurement when scrolling backward
2. Batching scroll adjustments via requestAnimationFrame

I’d strongly suggest splitting these into two separate PRs:

1. Backward scroll optimization
The option name reads like it disables scroll adjustment, but in reality it skips measurement under certain conditions. It might be worth either renaming it or being very explicit about this behavior in the docs.
There’s also an open question around how we guarantee that the layout eventually settles correctly after scrolling stops (i.e. when isScrolling flips back to false).

2. Scroll adjustment batching
There is no scroll adjustment batching implemented as by designed as when scrolling quickly, you see more visible white space. Because the scroll adjustments are delayed to the next requestAnimationFrame, the content height changes while the scroll offset isn’t corrected immediately, so during fast scrolling you get gaps until the batch flushes. The faster you scroll, the more noticeable this becomes.
This isn’t a trivial issue to solve, it’s the fundamental trade-off between less thrashing vs. more lag. With dynamic heights and fast scrolling, you really feel that lag.

dpwls02142:
> @dpwls02142 Thanks for the PR, this is addressing a very real problem. I think there are actually two distinct changes here:
> 
> 1. Skipping re-measurement when scrolling backward
> 2. Batching scroll adjustments via requestAnimationFrame
> 
> I’d strongly suggest splitting these into two separate PRs:
> 
> 1. Backward scroll optimization
>    The option name reads like it disables scroll adjustment, but in reality it skips measurement under certain conditions. It might be worth either renaming it or being very explicit about this behavior in the docs.
>    There’s also an open question around how we guarantee that the layout eventually settles correctly after scrolling stops (i.e. when isScrolling flips back to false).
> 2. Scroll adjustment batching
>    There is no scroll adjustment batching implemented as by designed as when scrolling quickly, you see more visible white space. Because the scroll adjustments are delayed to the next requestAnimationFrame, the content height changes while the scroll offset isn’t corrected immediately, so during fast scrolling you get gaps until the batch flushes. The faster you scroll, the more noticeable this becomes.
>    This isn’t a trivial issue to solve, it’s the fundamental trade-off between less thrashing vs. more lag. With dynamic heights and fast scrolling, you really feel that lag.

Hi @piecyk, thanks for the detailed feedback!

I agree with splitting this into two separate PRs. I'll focus on the backward scroll optimization and remove the batching approach.

**Changes**
- Removed all batching code (`pendingAdjustmentDeltas`, `scheduleAdjustmentBatch`, `flushAdjustments`, `adjustmentBatchTimer`)
- Renamed to `skipRemeasurementOnBackwardScroll` to clarify it's skipping measurement, not adjustment

**About layout stability**
You're right to ask about this. My thinking is that by the time a user scrolls backward, they've already scrolled forward past those items, so the measurements were already captured during the initial forward scroll. We're just reusing those cached values during the backward scroll to avoid the stuttering.

If an item's size genuinely changed while off-screen, it would get picked up the next time it enters the viewport during normal (non-backward) scrolling.

Let me know if you see any issues with this approach!

piecyk:
Closing this, as the same change was added in #1168

PR #1109, opened 2026-01-05
fix(angular-virtual): RuntimeError: NG0600: Writing to signals is not allowed in a computed or an effect by default

Stallion8472:
## 🎯 Changes

Fixes #1096 

## ✅ Checklist

- [x] I have followed the steps in the [Contributing guide](https://github.com/TanStack/virtual/blob/main/CONTRIBUTING.md).
- [x] I have tested this code locally with `pnpm run test:pr`.

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [ ] This change is docs/CI/dev-only (no release).

PR #1158, opened 2026-04-15
chore(angular-virtual): proper angular 19

benjavicente:
## 🎯 Changes

- Upgrades the adapter to Angular +19. Angular 19 is the latest angular version with LTS.
- Updates the implementation to use computed and linked signals instead of the lazy signal approach. This will make it possible to alias fields of the returned value, like `items = this.rowVirtualizer.getVirtualItems`. 
- Use vite as a bundler for angular since the package does not need any Angular compilation.

About #1096 and #1109, I couldn't replicate the issue with the latest version of virtual. With this upgrade to Angular 19, no effect could trigger directly set a value to a signal.

Pending:
- Confirmation that is ok to state that for Angular the adapter would only target [LTS version and up](https://angular.dev/reference/releases#actively-supported-versions), like query https://github.com/TanStack/query/issues/8707
- Aligning some of the following, like TS versions since Angular examples [need a newer one](https://angular.dev/reference/versions#actively-supported-versions).
- I did not manage to add unit tests due to tolling version alignment issues.

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/virtual/blob/main/CONTRIBUTING.md).
- [X] I have tested this code locally with `pnpm run test:pr`.

## 🚀 Release Impact

- [X] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **New Features**
  * Official Angular 19 support and signal-backed virtualizer state (including measurements cache, total size, virtual items, and scroll state).

* **Documentation**
  * Docs and README updated to require Angular 19+ and list available signal-backed virtualizer properties.

* **Chores**
  * Example projects and tooling bumped (Angular →19, TypeScript →5.6.3, Vite →6.4.2); package outputs moved to a dist/Vite layout; new e2e app and test config added.

* **Tests**
  * New Playwright E2E suites covering scrolling, smooth scroll, measurements, and stale-index scenarios.


piecyk:
@benjavicente awesome work! I’ll play around with it a bit. Could we add a simple e2e test here, similar to what we have in react-virtual, to cover the basic flow?

packages/angular-virtual/src/index.ts
```diff
@@ -27,55 +24,79 @@ export * from './types'
...
-
-  let cleanup: () => void | undefined
-  afterNextRender({ read: () => (virtualizer ?? lazyInit())._didMount() })
+  afterRenderEffect((cleanup) => {
+    cleanup(lazyVirtualizer()._didMount())
+  })
 
-  inject(DestroyRef).onDestroy(() => cleanup?.())
+  afterRenderEffect(() => {
+    reactiveVirtualizer()._willUpdate()
```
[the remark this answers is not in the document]
benjavicente: The change looks more aligned with react, where `_willUpdate` will be called on every re-render.

packages/angular-virtual/src/index.ts
```diff
@@ -27,55 +24,79 @@ export * from './types'
...
-      }
-    },
-    { allowSignalWrites: true },
-  )
-
-  let cleanup: () => void | undefined
-  afterNextRender({ read: () => (virtualizer ?? lazyInit())._didMount() })
+  afterRenderEffect((cleanup) => {
+    cleanup(lazyVirtualizer()._didMount())
+  })
```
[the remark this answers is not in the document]
benjavicente: `lazyVirtualizer()` does not change, so it runs once.

benjavicente:
All green now 🚀
I had to bump vite and typescript for type checks and sherif to to pass.

piecyk:
@benjavicente sorry about the conflicts! Could you sync with main and regenerate the lockfile? Once CI is green we can merge.

packages/angular-virtual/e2e/app/src/app/measure-element.component.ts
```diff
@@ -0,0 +1,116 @@
...
+        @for (row of virtualizer.getVirtualItems(); track itemKey(row.index)) {
+          <div
+            #virtualItem
+            [attr.data-testid]="itemKey(row.index)"
+            [attr.data-index]="row.index"
+            style="position: absolute; top: 0; left: 0; width: 100%;"
+            [style.transform]="'translateY(' + row.start + 'px)'"
+          >
+            <div style="display: flex; gap: 8px; align-items: center; padding: 4px;">
+              <span>Row {{ itemFor(row.index).label }}</span>
```
jsaguet: You could use `@let item = this.items()[$index]` and `item.id` to avoid itemFor/itemKey

packages/angular-virtual/src/index.ts
```diff
@@ -16,96 +16,157 @@ import {
...
+import { signalProxy } from './proxy'
+import type { ElementRef } from '@angular/core'
 import type { PartialKeys, VirtualizerOptions } from '@tanstack/virtual-core'
 import type { AngularVirtualizer } from './types'
 
 export * from '@tanstack/virtual-core'
 export * from './types'
 
-function createVirtualizerBase<
+export type AngularVirtualizerOptions<
```
jsaguet, line 30: It may be useful to add an injector option to pass to the 2 afterRenderEffects

.changeset/gold-days-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/angular-virtual': minor
```
piecyk: Let's use major as @angular/core goes from >=18.1.0 to >=19.0.0. Any consumer still on Angular 18 will break on install

```suggestion
'@tanstack/angular-virtual': major
```

packages/angular-virtual/src/proxy.ts
```diff
@@ -1,117 +1,110 @@
...
-          configurable: true,
-          enumerable: true,
-        })
+        return (target[property] = computed(
+          () => inputSignal()[property as keyof TInput],
+        ))
       }
 
-      return untypedTarget[property] || virtualizer[property as keyof V]
+      // All other fiels. Any fields that is not handled above will fail if the signal includes
```
piecyk:
```suggestion
      // All other fields. Any field that is not handled above will fail if the signal includes
```

packages/angular-virtual/src/index.ts
```diff
@@ -16,100 +19,181 @@ import {
...
-  )
-
-  let cleanup: (() => void) | undefined
-  afterNextRender({
-    read: () => {
-      cleanup = (virtualizer ?? lazyInit())._didMount()
-    },
-  })
+    })
+
```
piecyk, line 102: Let's add a comment   

// Lazy singleton: `computed` here is used purely for memoization, the body
// reads `resolvedOptions` via `untracked`, so it has no reactive deps and the
// `Virtualizer` instance is constructed exactly once, on first access.

packages/angular-virtual/src/index.ts
```diff
@@ -16,100 +19,181 @@ import {
...
+    const lazyVirtualizer = computed(
+      () => new Virtualizer(untracked(resolvedOptions)),
+    )
+
```
piecyk, line 109: // Note: `reactiveVirtualizer` is referenced by the `onChange` closure inside
// `resolvedOptions` above. That's safe because `onChange` is only invoked by
// virtual-core observers after this `linkedSignal` is initialized, but be
// aware of the TDZ when refactoring the order of these declarations.

piecyk (changes requested):
Let’s bump this to a major version and ship it.

piecyk (approved):

PR #1159, opened 2026-04-16
fix(angular-virtual): capture _didMount cleanup to remove scroll/resize listeners on destroy

jsaguet:
## 🎯 Changes

`afterNextRender` discarded the teardown function returned by _didMount(), so observeWindowOffset's scroll and resize listeners were never unregistered.

Also fixes the cleanup variable type annotation.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where scroll and resize event listeners were not being properly removed when the Angular virtual component is destroyed, preventing potential memory leaks.

### DIFF
--- a/.changeset/fix-angular-virtual-cleanup.md
+++ b/.changeset/fix-angular-virtual-cleanup.md
@@ -2,4 +2,4 @@
 '@tanstack/angular-virtual': patch
 ---
 
-fix: capture _didMount cleanup to remove scroll/resize listeners on destroy
+fix: capture \_didMount cleanup to remove scroll/resize listeners on destroy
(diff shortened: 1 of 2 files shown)


piecyk:
I think #1158 may cover this via the Angular 19 afterRenderEffect lifecycle, which looks like a cleaner way to register _didMount() cleanup than the current afterNextRender + DestroyRef pattern. If that lands as-is, this may no longer be needed separately.

jsaguet:
I completely agree with you that #1158 would be much better. 
But since it's still a draft, I thought this fix was still worth it in the meantime considering that every app using this lib currently faces these leaks and errors caused by this lack of cleanup.

Also, this fix would be available for v18 apps where #1158 would drop support for it.

piecyk:
@jsaguet Good point 👍

piecyk:
Thanks @jsaguet, merging now! Btw if you get a chance, a review on #1158 would be much appreciated.

Issue #884, opened 2024-11-22
Window & Dynamic Size List Virtualization Causes Mobile Safari IOS Scroll Momentum Interruption


kevin-sjj:
### Describe the bug

By using the **useWindowVirtualizer** hook to apply virtualization for list items with **different heights** in relation to window scrolling, when you scroll, the virtualization behavior will occur.

Items that no longer need to be rendered will disappear from the DOM, and the list items corresponding to the visible indices will begin to appear in the DOM.

On **Mobile iOS Safari**, momentum scrolling triggered by user gestures works, but when combined with virtualization, the momentum effect fails to end smoothly and instead stutters or breaks abruptly. Is there a way to resolve this issue?

```tsx
import "./styles.css";
import { useWindowVirtualizer } from "@tanstack/react-virtual";

const createRandomHeights = (
  min: number,
  max: number,
  count: number = 1000
) => {
  const heights = [];
  for (let i = 0; i < count; i++) {
    heights.push(Math.floor(Math.random() * (max - min + 1) + min));
  }
  return heights;
};

const randomHeights = createRandomHeights(100, 400);
const colors = ["red", "yellow", "blue"];

export default function App() {
  const virtualizer = useWindowVirtualizer({
    count: 1000,
    estimateSize: () => 400,
    overscan: 5,
    scrollMargin: 0,
  });

  const virtualItems = virtualizer.getVirtualItems();

  return (
    <div>
      <div
        style={{
          height: `${virtualizer.getTotalSize()}px`,
          position: "relative",
          width: "100%",
        }}
      >
        <div
          style={{
            transform: `translateY(${
              (virtualItems[0]?.start ?? 0) - virtualizer.options.scrollMargin
            }px)`,
            position: "absolute",
            left: "0",
            top: "0",
            width: "100%",
          }}
        >
          {virtualItems.map((virtualItem) => {
            return (
              <div
                key={virtualItem.key}
                data-index={virtualItem.index}
                ref={virtualizer.measureElement}
                style={{
                  paddingTop: "40px",
                }}
              >
                <div
                  style={{
                    height: randomHeights[virtualItem.index],
                    backgroundColor: colors[virtualItem.index % 3],
                  }}
                ></div>
              </div>
            );
          })}
        </div>
      </div>
    </div>
  );
}

```

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/qhkvyz

### Steps to reproduce

1. Go to CodeSandbox Link
[...]

PR #1189, opened 2026-06-08
feat(virtual-core): iOS momentum-safe scroll adjustments via CSS offset

piecyk:
On iOS WebKit, writing `scrollTop` during momentum scroll cancels the in-flight scroll. Instead of writing `scrollTop`, we now apply a negative `marginTop` on the container element to visually compensate for above-viewport size changes. The CSS offset is flushed to a real `scrollTop` write once momentum fully settles.

- Defer scroll adjustments during iOS touch and momentum phases using CSS offset (`marginTop`/`marginLeft`)
- Force-flush CSS offset before programmatic scroll operations (`scrollToIndex`, `scrollToOffset`, `scrollBy`)
- Compensate `scrollOffset` for active CSS offset in range calculations
- Guard against Safari elastic overscroll (rubber-band) during flush
- Clean up CSS offset on unmount
- Refine backward-scroll suppression: first measurements always adjust regardless of direction; re-measurements skip during backward scroll to avoid the `scrollTop` cascade jank

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Improvements**
  * Smoother iOS Safari momentum scrolling via deferred CSS-based offset compensation and safer flush handling.
  * Programmatic scrolling now forces resolution of any active iOS deferred offsets before performing scrolls.

* **Behavior**
  * Refined resize-scroll adjustments: always apply the first backward-scroll measurement, skip redundant re-measurement adjustments, and keep offset calculations consistent during deferral.

* **Documentation**
  * Chat/virtualization guidance now requires stable item keys and recommends disabling scroll anchoring.

* **Tests**
  * Updated tests to match refined forward/backward resize-adjustment behavior.

### DIFF
--- a/.changeset/ios-momentum-scroll.md
+++ b/.changeset/ios-momentum-scroll.md
@@ -1,0 +1,14 @@
+---
+'@tanstack/virtual-core': minor
+---
+
+feat: iOS momentum-safe scroll adjustments via CSS offset
+
+On iOS WebKit, writing `scrollTop` during momentum scroll cancels the in-flight scroll. Instead of writing `scrollTop`, we now apply a negative `marginTop` on the container element to visually compensate for above-viewport size changes. The CSS offset is flushed to a real `scrollTop` write once momentum fully settles.
+
+- Defer scroll adjustments during iOS touch and momentum phases using CSS offset (`marginTop`/`marginLeft`)
+- Force-flush CSS offset before programmatic scroll operations (`scrollToIndex`, `scrollToOffset`, `scrollBy`)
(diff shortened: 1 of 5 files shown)


piecyk:
Closing as superseded. The deferral approach that landed on `main` covers #884 without the CSS `marginTop` offset, and #1280 fixes its remaining over-reach (#1250) by gating the deferral on touch provenance rather than `isScrolling`. The rubber-band guard and the force-flush-before-programmatic-scroll ideas from this branch both live on in that path.

PR #1193, opened 2026-06-12
perf: reduce per-scroll and per-render allocations

mds-ant:
## Summary

A pass over `virtual-core` and `react-virtual` to cut per-scroll-frame and per-render allocations. Each optimization is its own commit so they can be reviewed (or cherry-picked) independently.

Builds on the `getItemKey` memo-dep fix in the first commit on this branch (`d63ac20`), which is the largest single win — it stops an inline `getItemKey` closure from forcing a full O(n) measurements rebuild on every render.

## Benchmarks

`packages/virtual-core/tests/bench.bench.ts` (added a scroll-loop section in this PR). hz = ops/sec, higher is better. No benchmark regressed below 0.97×.

| Benchmark | Before | After | Speedup |
|---|--:|--:|--:|
| `setOptions × 10,000` | 309.6 | 633.0 | **2.04×** |
| Scroll loop, n=1k, 10k scrolls (`calculateRange`) | 104.4 | 168.3 | **1.61×** |
| Scroll loop, n=100k, 10k scrolls | 80.3 | 128.4 | **1.60×** |
| `getVirtualItemForOffset(0)`, n=1k | 54,997 | 82,243 | **1.50×** |
| Cold mount `getMeasurements`, n=500k | 154.3 | 224.0 | **1.45×** |
| onChange storm, n=100 | 83,696 | 120,179 | **1.44×** |
| Cold mount `getMeasurements`, n=100k | 823.2 | 1,106.0 | **1.34×** |
| Scroll + `getVirtualItems`, n=1k | 39.5 | 52.1 | **1.32×** |
| Scroll + `getVirtualItems`, n=100k | 30.3 | 37.2 | **1.23×** |

## Commits

| Commit | Change |
|---|---|
| `bench(core)` | Add scroll-loop benchmark — none of the existing benches covered the per-scroll-frame path |
| `perf(core): indexed for-loop dep compare in memo()` | `.some()` allocates a closure on every memo invocation; memo runs ~4×/scroll, ~6×/render |
| `perf(core): hoist setOptions nested-default allocations` | `() => {}` / `{w:0,h:0}` / `[]` → module constants |
| `perf(core): inline zero-arg debounce in observeOffset` | Exported `debounce` allocates rest-args + setTimeout closure per scroll; only internal caller passes zero args. Public API unchanged |
| `perf(core): replace getMeasurementOptions memo with version int` | Was a memo-hop allocating a 6-tuple on every `getMeasurements()`; `setOptions` already has `prevOptions` in hand |
| `perf(core): persist lazy-view item cache across rebuilds` | `createLazyMeasurementsView` discarded `new Array(count)` on every rebuild; now invalidates only the dirtied tail so VirtualItems at indices < `pendingMin` keep referential identity (lets row `React.memo` bail out) |
| `perf(core): zero-alloc calculateRange on the lanes===1 hot path` | Positional args, monomorphic `findNearestBinarySearchFlat`, mutate `this.range` in place |
| `perf(core): batch ResizeObserver entries into one notify` | Was N× `notify` per RO batch; now process all entries, fire once. Also lifts rAF outside the loop |
| `perf(react-virtual): stable onChange + reuse adapter options object` | Stable `onChange` via latest-ref, copy fields onto a persisted options object instead of `{...spread}` per render, hoist `useWindowVirtualizer` defaults |

## Behavior change

Only one (from the `getItemKey` commit): if `getItemKey` returns *different keys for the same indices* without `count` changing, consumers must call `virtualizer.measure()` to rebuild — matching the existing `estimateSize` contract. The `anchorTo: 'end'` path is unaffected (it already detects edge-key changes in `setOptions`).

## Tests

- 94/94 existing `virtual-core` tests pass
- 3 new tests in `tests/getItemKey-identity.test.ts` (fail on `main`, pass with this branch)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized scroll event handling to reduce rendering overhead and improve responsiveness during fast scrolling scenarios.
  * Enhanced memory efficiency through improved caching and item measurement strategies.
  * Reduced unnecessary re-renders by stabilizing internal callbacks and state management.

### DIFF
--- a/packages/react-virtual/src/index.tsx
+++ b/packages/react-virtual/src/index.tsx
@@ -72,22 +72,24 @@ export type ReactVirtualizerOptions<
 function useVirtualizerBase<
   TScrollElement extends Element | Window,
   TItemElement extends Element,
->({
-  useFlushSync = true,
-  directDomUpdates = false,
-  directDomUpdatesMode = 'transform',
-  ...options
-}: ReactVirtualizerOptions<TScrollElement, TItemElement>): ReactVirtualizer<
-  TScrollElement,
-  TItemElement
-> {
+>(
+  rawOptions: ReactVirtualizerOptions<TScrollElement, TItemElement>,
+): ReactVirtualizer<TScrollElement, TItemElement> {
+  const {
+    useFlushSync = true,
+    directDomUpdates = false,
+    directDomUpdatesMode = 'transform',
+  } = rawOptions
   const rerender = React.useReducer((x: number) => x + 1, 0)[1]
 
-  // Mutable across renders so the onChange closure captured by setOptions
-  // always reads the latest values without us having to re-create it.
-  const directRef = React.useRef({
+  // Mutable across renders so the onChange closure (created once, in the
+  // useState initializer below) always reads the latest values without us
+  // having to re-create it.
+  const latestRef = React.useRef({
     enabled: directDomUpdates,
     mode: directDomUpdatesMode,
+    useFlushSync,
+    userOnChange: rawOptions.onChange,
     container: null as HTMLElement | null,
     lastSize: null as number | null,
     // Keyed by the element itself so a remounted node (same key, new DOM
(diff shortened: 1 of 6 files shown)


piecyk:
Wow, this looks really interesting! Why was this closed?

Is any help needed to get this into a shape that could land, maybe by splitting it up or adding more validation? Would love to help make this blazing fast.

PR #1205, opened 2026-06-21
feat(benchmarks): add React Aria Virtualizer to cross-library harness

Stanzilla:
## 🎯 Changes

Add React Aria Components to the reproducible Playwright benchmark matrix:

- `rac` — `Virtualizer` + `ListBox` (full integrated stack)
- `tanstack-rac` — TanStack virtual + WAI-ARIA `listbox`/`option` roles (no RAC collection)
- `rac-listbox` — `ListBox` without virtualization (collection overhead baseline)

Also improves the shared harness (`getSearchRoot`, `waitForTargetItem`) and consolidates per-library scenario exclusions in `libScenarioExclusions.mjs`.

### Why `rac-listbox` excludes `mount-fixed-100k`

`rac-listbox` deliberately does **not** use React Aria's `Virtualizer` — it renders every `ListBoxItem` in the DOM so we can measure collection/selection overhead in isolation. That is useful at 1k–10k, but at **100k items it mounts 100k React nodes**, which is not a realistic workload for a non-virtualized list:

- On the author's machine, **10k already uses ~534 MB heap** (vs ~3–14 MB for virtualized libraries).
- A 100k run would likely OOM or hang Chromium/Playwright before producing meaningful numbers.
- The scenario exists to compare **virtualizer** mount cost at scale; including a non-virtualized baseline there would be a stress test of the browser, not a fair library comparison.

So `LIB_SCENARIO_EXCLUSIONS` skips `mount-fixed-100k` for `rac-listbox` only. All other scenarios still run — including accuracy probes, where `rac-listbox` is actually the most reliable baseline because every row stays mounted.

Latest RAC numbers (2 runs per cell) are documented in `benchmarks/README.md`.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm run test:pr` (benchmark build + `node runner/run.mjs` subset).

## 🚀 Release Impact

- [ ] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [x] This change is docs/CI/dev-only (no release).

## Summary by CodeRabbit

* **New Features**
  * Added benchmark support for React Aria Components (RAC) and TanStack RAC variants, including new RAC listbox-style pages.
  * Expanded “Library matrix” coverage with updated results tables and comparisons.
* **Bug Fixes**
  * Corrected routing for the Virtuoso benchmark page.
* **Documentation**
  * Clarified benchmark methodology for RAC accuracy, including handling when targets mount after scrolling.
  * Added guidance for excluding unsupported scenarios per library.
* **Other Changes**
  * Updated benchmark execution to automatically skip excluded scenario/library combinations.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -191,7 +191,7 @@ Add an entry to `SCENARIOS` in `src/scenarios/types.ts`. The runner discovers it
 1. Create `src/pages/MyLibPage.tsx` that registers a `HarnessHandle` (see existing pages for the contract).
 2. Wire it into `src/main.tsx`'s switch.
 3. Add the library name to `ALL_LIBS` in `runner/run.mjs`.
-4. If a library cannot run certain scenarios, add exclusions to `LIB_SCENARIO_EXCLUSIONS` in `runner/run.mjs` and `scenarios/types.ts`.
+4. If a library cannot run certain scenarios, add exclusions to `LIB_SCENARIO_EXCLUSIONS` in `src/scenarios/libScenarioExclusions.mjs` (also re-exported from `scenarios/types.ts`).
 
 ## Known limitations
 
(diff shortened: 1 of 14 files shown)


Stanzilla:
@devongovett heads up, can you check if this looks fair to you?

PR #1206, opened 2026-06-22
perf(virtual-core): drop closures from `calculateRange` on the `lanes===1` hot path

mds-ant:
## 🎯 Changes

Cut per-scroll-frame allocations on the default `lanes === 1` path. Range computation previously allocated an options object and two closures on every scroll event; it now does the same work allocation-free, reducing GC pressure during continuous scroll.

- `calculateRangeImpl` takes positional args instead of an options object.
- New `findNearestBinarySearchFlat(flat, high, value)` — monomorphic typed-array binary search that indexes `flat[i*2]` directly. The closure-based `findNearestBinarySearch` is kept for the `lanes > 1` / no-`flat` fallback.
- The `lanes === 1` forward-walk reads `start + size` from the typed array inline.

### Motivation

`scrollOffset` is a memo dep on `calculateRange`, so its body runs on every scroll event. On `main` each run …

- allocates a 5-field `{measurements, outerSize, scrollOffset, lanes, flat}` options object,
- allocates two `getStart` / `getEnd` closures, and
- pays a polymorphic indirect call through those closures for each binary-search probe.

The codebase already has a `Float64Array` fast path for `lanes === 1` (`_flatMeasurements`). This change reads from it directly instead of through closures.

### Performance

| Benchmark | `main` (Hz) | This PR (Hz) | Δ |
|---|--:|--:|--:|
| `n=1000, 10k scrolls` | 104.7 | 130.7 | +25% |
| `n=100000, 10k scrolls` | 87.7 | 109.1 | +24% |
| `n=1000, 10k scrolls + getVirtualItems` | 34.0 | 37.1 | +9% |
| `n=100000, 10k scrolls + getVirtualItems` | 32.5 | 35.3 | +9% |

## ✅ Checklist

## 🚀 Release Impact

This PR was assisted by Claude Code.

## Summary of Changes

* **Bug Fixes**
  * Clears virtualized range state when there are no measurements or when the outer container size is zero.

* **Performance Improvements**
  * Reduced per-scroll allocations during continuous scrolling, with the biggest gains when running with a single lane.

* **Tests**
  * Added a new scroll-loop benchmark suite that simulates thousands of rapid scroll frames to compare range-only vs. full virtual item updates.

### DIFF
--- a/.changeset/calculate-range-fewer-allocs.md
+++ b/.changeset/calculate-range-fewer-allocs.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Cut per-scroll-frame allocations on the default `lanes === 1` path. Range computation previously allocated an options object and two closures on every scroll event; it now does the same work allocation-free, reducing GC pressure during continuous scroll.
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -1334,22 +1334,22 @@ export class Virtualizer<
       this.options.lanes,
     ],
     (measurements, outerSize, scrollOffset, lanes) => {
-      return (this.range =
-        measurements.length > 0 && outerSize > 0
-          ? calculateRange({
-              measurements,
-              outerSize,
-              scrollOffset,
-              lanes,
-              // Pass the typed array so binary search + forward-walk can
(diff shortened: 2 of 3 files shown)


piecyk:
Thanks @mds-ant 👏

PR #1208, opened 2026-06-23
Scrolling fix

dracofulmen:
Added check to observeElementOffset to prevent continuous re-renders

## 🎯 Changes

Currently, `observeElementOffset` causes `isScrolling` to get stuck as true for as long as `offset > 0`. This causes continuous rerenders even after scrolling stops. The `this.maybeNotify();` call also causes continuous rerenders while scrolling. Both of these issues cause the scroll bar to flicker. This fix adds a check to make sure that scrolling only causes notifications once, when the scrolling stops.

## ✅ Checklist

- [✅] I have followed the steps in the [Contributing guide](https://github.com/TanStack/virtual/blob/main/CONTRIBUTING.md).
- [✅] I have tested this code locally with `pnpm run test:pr`.

## 🚀 Release Impact

- [✅] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a continuous re-render issue during or after scrolling in the virtualized list.
  * Improved scroll offset observation by skipping redundant intended-offset reconciliation when the reported offset already matches the current scroll position during active scrolling.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -779,6 +779,13 @@ export class Virtualizer<
           // self-write — by the time the user has moved 1.5 px, the
           // intended value will already have been consumed by a prior
           // scroll event and cleared.
+          if (offset === this.scrollOffset && isScrolling === true) {
+            // this prevents continous rerenders
+            isScrolling = false;
+            this.isScrolling = false;
+            this.scrollOffset = offset
+            this.maybeNotify();
+          }
           if (
             this._intendedScrollOffset !== null &&
             Math.abs(offset - this._intendedScrollOffset) < 1.5
@@ -804,7 +811,6 @@ export class Virtualizer<
           if (this.scrollState) {
             this.scheduleScrollReconcile()
           }
-          this.maybeNotify()
         }),
       )
 
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -779,11 +779,7 @@ export class Virtualizer<
           // which forces a render that triggers another such event: an
           // infinite re-render loop. Ignore it. (Self-writes are handled by
           // the `_intendedScrollOffset` reconciliation just below.)
-          if (
-            isScrolling &&
-            this._intendedScrollOffset === null &&
-            offset === this.scrollOffset
-          ) {
+          if (isScrolling && offset === this.scrollOffset) {
             return
           }
           
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -811,6 +811,7 @@ export class Virtualizer<
           if (this.scrollState) {
             this.scheduleScrollReconcile()
           }
+          this.maybeNotify()
         }),
       )
 
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -794,13 +794,6 @@ export class Virtualizer<
           // self-write — by the time the user has moved 1.5 px, the
           // intended value will already have been consumed by a prior
           // scroll event and cleared.
-          if (offset === this.scrollOffset && isScrolling === true) {
-            // this prevents continous rerenders
-            isScrolling = false;
-            this.isScrolling = false;
-            this.scrollOffset = offset
-            this.maybeNotify();
-          }
           if (
             this._intendedScrollOffset !== null &&
             Math.abs(offset - this._intendedScrollOffset) < 1.5
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -779,7 +779,11 @@ export class Virtualizer<
           // which forces a render that triggers another such event: an
           // infinite re-render loop. Ignore it. (Self-writes are handled by
           // the `_intendedScrollOffset` reconciliation just below.)
-          if (isScrolling && offset === this.scrollOffset) {
+          if (
+            isScrolling &&
+            this._intendedScrollOffset === null &&
+            offset === this.scrollOffset
+          ) {
             return
           }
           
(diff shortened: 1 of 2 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -779,6 +779,13 @@ export class Virtualizer<
           // self-write — by the time the user has moved 1.5 px, the
           // intended value will already have been consumed by a prior
           // scroll event and cleared.
+          if (offset === this.scrollOffset && isScrolling === true) {
+            // this prevents continous rerenders
+            isScrolling = false;
+            this.isScrolling = false;
+            this.scrollOffset = offset
+            this.maybeNotify();
+          }
```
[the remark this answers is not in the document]
dracofulmen: The problem with this is that calling this.maybeNotify() while scrolling causes part of the issue. It means that the component continuously rerenders while scrolling, causing the scroll bar to flicker.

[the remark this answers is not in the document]
dracofulmen: This still leads to rerenders while scrolling. Fundamentally, any call to this.maybeNotify() will cause a rerender, so that should only happen when absolutely necessary.

piecyk (commented):
@dracofulmen Hi, could you clarify what "continuous re-renders" means concretely here, what's measured as re-rendering, and under what interaction?

Basic maybeNotify is memoized on [isScrolling, startIndex, endIndex], so a scroll event at an unchanged offset can't cause a re-render unless isScrolling toggles (the range can't have moved)

dracofulmen:
@piecyk I'm using `@workday/canvas-kit-react` v15.0.14. Here's the video of the issue: [Reproduction.mov](https://github.com/user-attachments/assets/3a94b2bf-a5e3-444f-8987-4843fbd4f901). And here's the code that I used to reproduce it: [minimal.tsx](https://github.com/user-attachments/files/29219845/minimal.tsx). When I dove into the react devtools, I found that `isScrolling` gets stuck on true as long as the scroll bar isn't at the top. This is why I added the `offset === this.scrollOffset && isScrolling === true` check. However, without moving the `this.maybeNotify()` call, I still got the flashes during scrolling.

piecyk:
> @piecyk I'm using `@workday/canvas-kit-react` v15.0.14. Here's the video of the issue: [Reproduction.mov](https://github.com/user-attachments/assets/3a94b2bf-a5e3-444f-8987-4843fbd4f901). And here's the code that I used to reproduce it: [minimal.tsx](https://github.com/user-attachments/files/29219845/minimal.tsx). When I dove into the react devtools, I found that `isScrolling` gets stuck on true as long as the scroll bar isn't at the top. This is why I added the `offset === this.scrollOffset && isScrolling === true` check. However, without moving the `this.maybeNotify()` call, I still got the flashes during scrolling.

Hmm, that looks off. Could you prepare a reproducible example on stackblitz using the minimal example you mentioned, so we can investigate further?

Once we have that, we can dig into why isScrolling remains stuck.

dracofulmen:
@piecyk Here's the [stackblitz reproduction](https://stackblitz.com/edit/vitejs-vite-kjxpnqnw?file=src%2Fminimal.tsx)

piecyk:
> @piecyk Here's the [stackblitz reproduction](https://stackblitz.com/edit/vitejs-vite-kjxpnqnw?file=src%2Fminimal.tsx)

@dracofulmen thanks! Could you share what environment you’re seeing this issue in?

I tested it on Chrome and Firefox on macOS and it looks good, no flickering on my side.

Also, virtualization is not enabled by default. You need to pass shouldVirtualize: true to useSelectModel to enable it.

dracofulmen:
I'm using it in Firefox on Mac.

packages/virtual-core/src/index.ts
piecyk: Let's move this check up
```suggestion
```

packages/virtual-core/src/index.ts
piecyk, line 790:
```suggestion
          // A scroll event that reports movement but lands on the offset we
          // already hold — and isn't a self-write read-back — is a spurious
          // no-op re-emit that Safari/Firefox fire after a re-render's layout
          // (Chrome doesn't). Treating it as scrolling re-arms `isScrolling`,
          // which forces a render that triggers another such event: an
          // infinite re-render loop. Ignore it. (Self-writes are handled by
          // the `_intendedScrollOffset` reconciliation just below.)
          if (
            isScrolling &&
            this._intendedScrollOffset === null &&
            offset === this.scrollOffset
          ) {
            return
          }
          
          // If this scroll event looks like the browser's read-back of a
```

packages/virtual-core/src/index.ts
```diff
@@ -804,7 +811,6 @@ export class Virtualizer<
           if (this.scrollState) {
             this.scheduleScrollReconcile()
           }
```
piecyk, line 821: Let;s revert this line
```suggestion
          }
          this.maybeNotify()
```

piecyk (changes requested):
Thanks @dracofulmen, the re-render loop is real. I can reproduce it reliably in Firefox, though not in Chrome, with the virtualized Select.

After scrolling stops, it keeps re-rendering indefinitely, roughly every ~168ms, which lines up with isScrollingResetDelay.

Bit tweaked your fix to ignore a scroll event when both of these are true:
* the event lands on the offset we already have,
* it is not a self-write read-back.

Real scrolls still change the offset, so they pass through normally. Programmatic scrolls are still handled by the _intendedScrollOffset reconciliation below.

I tested it, and the loop is gone while scrolling still works.

One thing I want to be upfront about: I did not isolate exactly why the browser fires that extra scroll event. I know it is browser-generated and happens right after each render, but I have not pinned down the underlying mechanism.

It would be nice to move @workday/canvas-kit-react to directDomUpdates. That should reduce the number of re-renders while scrolling:
https://tanstack.com/virtual/latest/docs/framework/react/react-virtual#directdomupdates

Feel free to create a PR on their side and ping me for review.

piecyk (approved):

PR #1209, opened 2026-06-25
fix(virtual-core): sync `scrollOffset` in `applyScrollAdjustment` so end-anchored resize is not lost to browser clamp

mds-ant:
## 🎯 Changes

With `anchorTo: 'end'` and a dynamically growing last item (e.g. a streaming chat reply measured via `measureElement`), the viewport drifts away from the end starting on the **second** resize tick.

`resizeItem`'s `wasAtEnd` branch runs in this order:

1. update `itemSizeCache` → `getTotalSize()` is now larger
2. `applyScrollAdjustment(newTotal − prevTotal)` → `_scrollToOffset(getScrollOffset(), {adjustments: delta})` → `el.scrollTo({top: scrollOffset + delta})`
3. `notify(false)` → consumer re-renders → sets the sizer `style.height` to the new `getTotalSize()`

Step 2 writes `scrollTop` while the container is still at the **old** total height (the consumer hasn't re-rendered yet), so the browser clamps the write back to the old max. Because `scrollTop` didn't move, no scroll event fires, `this.scrollOffset` stays stale, and the next tick's `getVirtualDistanceFromEnd()` reads `newTotal − viewport − staleOffset > scrollEndThreshold` → `wasAtEnd = false` → no further adjustments.

**Fix:** in `applyScrollAdjustment`, after `_scrollToOffset` has recorded `_intendedScrollOffset`, also bake the adjustment into `this.scrollOffset` and zero `scrollAdjustments` (so their sum stays invariant for callers like the `itemStart < getScrollOffset() + scrollAdjustments` predicate). The DOM write may still be clamped, but `scrollOffset` now carries the intended position so the next `wasAtEnd` check stays true and the adjustment lands once the consumer has grown the sizer. This is the same idea as #1176 (eager `scrollOffset` adjustment for prepend in `setOptions`), applied to the resize-adjust path.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/resize-adjust-scrolloffset-sync.md
+++ b/.changeset/resize-adjust-scrolloffset-sync.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Sync `scrollOffset` in `applyScrollAdjustment` so end-anchored streaming resize isn't lost to browser clamp
+
+With `anchorTo: 'end'` and a dynamically growing last item (token streaming), `resizeItem` writes the scroll adjustment to `scrollTop` before the consumer has grown the sizer, so the browser clamps the write and no scroll event fires. `scrollOffset` stayed stale, the next tick's `wasAtEnd` check failed, and the viewport drifted away from the end. This fix carries the intended target in `scrollOffset` (zeroing `scrollAdjustments`) the same way the prepend path in `setOptions` does, so the next `getVirtualDistanceFromEnd()` reads the post-adjustment position.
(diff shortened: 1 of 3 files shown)


mds-ant:
@piecyk Does this look good to you?

piecyk (approved):
Nice fix 👏 Let's 🚢 🇮🇹 

Just a side note, and non-blocking: the iOS path isn’t covered here. The isIOSWebKit() && scrolling branch defers and never reaches the new code, so the same drift could still happen there.

There are a few broader issues around isIOSWebKit, so I’ll add this there.

piecyk:
Thanks @mds-ant

PR #1212, opened 2026-06-30
fix(virtual-core): viewport drifts when above-viewport rows resize over multiple frames

mds-ant:
## 🎯 Changes

Folding adjustments eagerly into `scrollOffset` means the browser's scroll event for an adjustment write now arrives at exactly the offset we already hold, and the direction computation classified that equality as `'backward'`. With `isScrolling` held for `isScrollingResetDelay` after the event, the default `shouldAdjustScrollPositionOnItemSizeChange` then skipped every above-viewport re-measure compensation until the reset — so multi-frame reflows (a side pane's width animation re-wrapping rows while the user is scrolled up) drifted the viewport, with only ~1 frame in 9 compensated.

An event at the held offset carries no directional information, so keep the previous direction for it instead of computing the comparison. Real gestures still latch normally (their read-backs never equal the held offset, which moves with each event), an equality read-back during a real backward gesture correctly stays `'backward'`, and the `isScrolling === false` reset emission still clears the direction.

## ✅ Checklist

## 🚀 Release Impact

This PR was assisted by Claude Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scroll direction handling so the app no longer misreads its own scroll adjustment read-backs as user scrolling.
  * Prevents viewport drift and skipped compensation during multi-frame reflows, especially after content changes above the viewport.
  * Real scroll gestures still update direction normally, and direction now stays unchanged when the offset hasn’t actually moved.

### DIFF
--- a/.changeset/scroll-direction-self-write-readback.md
+++ b/.changeset/scroll-direction-self-write-readback.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Don't latch a scroll direction from the read-back of the virtualizer's own adjustment write
+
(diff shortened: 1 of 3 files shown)


mds-ant:
@piecyk, this is a follow-up to https://github.com/TanStack/virtual/pull/1209.

PR #1220, opened 2026-07-06
fix(virtual-core): reset iOS gesture/deferral state on cleanup

basnijholt:
`cleanup()` resets `scrollState`, `scrollElement`, and `targetWindow`, but not the iOS gesture/deferral state: `_iosDeferredAdjustment`, `_iosTouching`, `_iosJustTouchEnded`.

That state is scoped to the current scroll element. The touch listeners maintaining it are removed by the unsubs, and an in-flight touch keeps targeting the old element (implicit touch capture), so the new element never delivers the matching `touchend`. Carrying the state across a scroll-element swap breaks three ways:

- A deferred adjustment accumulated against the old element is applied to the new element's `scrollTop` on its first flush.
- A swap mid-touch strands `_iosTouching = true`. Every adjustment on the new element is then deferred, and `_flushIosDeferredIfReady` early-returns, until the user's next full touch cycle.
- A swap inside the post-touchend grace window strands `_iosJustTouchEnded = true` indefinitely: the listener unsub clears the grace timer, and that timer held the only pending reset of the flag.

This is easy to hit in chat UIs, where switching conversations remounts the scroller while the user is mid-flick.

Fix: reset the three fields in `cleanup()`. The grace timer needs no extra handling since the listener unsub already clears it.

Tests: three regression tests following the existing Phase 1 patterns, one per failure mode. On current `main` without the src change they fail (3 failed / 102 passed); with it, 105 pass. `pnpm vitest run`, `tsc`, and `eslint ./src` in `packages/virtual-core` are green.

Found while auditing the 3.2.0 to 3.14.5 upgrade in a Matrix client fork (mindroom-ai/mindroom-cinny#83).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where iOS touch and scroll state could persist when switching scroll elements.
  * Prevented deferred scroll adjustments from being incorrectly applied to a newly assigned element.
  * Ensured post-touch interaction state and related timers are cleared during cleanup.

* **Chores**
  * Added a patch release for the virtual-core package.

### DIFF
--- a/.changeset/reset-ios-deferral-state-on-cleanup.md
+++ b/.changeset/reset-ios-deferral-state-on-cleanup.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Reset iOS gesture/deferral state in `cleanup()` so it no longer leaks across scroll element swaps.
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -741,6 +741,19 @@ export class Virtualizer<
       this.rafId = null
     }
     this.scrollState = null
+    // The iOS gesture/deferral state is scoped to the current scroll
+    // element: the touch listeners that maintain it were just removed, and
+    // an in-flight touch keeps targeting the old element (implicit touch
+    // capture), so the new element never reports it. Carrying the state
+    // over would replay a stale deferred delta on the new element's first
+    // flush, and a cleanup that lands mid-touch or inside the post-touchend
+    // grace window would strand _iosTouching / _iosJustTouchEnded as true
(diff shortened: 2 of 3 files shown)


piecyk (commented):
@basnijholt Looks good! Could you also add a patch changeset for @tanstack/virtual-core before we merge this?

basnijholt:
Done, added a patch changeset for `@tanstack/virtual-core` in 2ff311c.

piecyk:
> Done, added a patch changeset for `@tanstack/virtual-core` in [2ff311c](https://github.com/TanStack/virtual/commit/2ff311c0e4e04fdaa3f4bc8286f5a2b316ccd433).

Awesome, thanks!

PR #1224, opened 2026-07-08
perf(virtual-core): replace multi-lane backward scan with incremental laneEnds argmin

2wheeh:
## 🎯 Changes

Multi-lane (`lanes > 1`) placement in `getMeasurements` found the shortest lane by scanning `measurements` backward through already-placed items (`getFurthestMeasurement`) on every uncached placement. This PR keeps a running `laneEnds` array, updated in the build loop, and picks the shortest lane with an `O(lanes)` argmin instead — so per-placement work no longer grows with scan depth. `getMeasurements` is ~3–11× faster on multi-lane cold mount and rebuilds.

- `getMeasurements` keeps a `laneEnds: Float64Array(lanes)` and a `filledLanes` counter, updated as each item is placed.
- The uncached branch picks the shortest lane via `argmin(laneEnds)`, keeping the tie-break (`end`, then last-item index) exact so placement is unchanged.
- `getFurthestMeasurement` is removed; the `lanes === 1` fast path is untouched.

> Sibling of #1206, which dropped closures from `calculateRange` on the `lanes === 1` hot path. This one targets the `lanes > 1` placement path.

### Correctness

Placement is byte-for-byte identical to the old algorithm for the supported `gap >= 0` range. `tests/lane-equivalence.test.ts` ports the original backward-scan verbatim as an oracle and asserts identical `lane` / `start` / `size` / `end` for every item — across fresh `estimate`-mode builds (lanes ∈ {2,3,4,5,8} × 20 seeds) and `measured`-mode rebuilds driven by random `resizeItem()` (covering incremental `min > 0` rebuilds and the cached-lane path).

> [!NOTE]
> Negative `gap`: the old scan tracked each lane's *max-end* item, while `laneEnds` tracks its *last* item — these coincide for any `gap >= 0`. A negative `gap` that overlaps items can shift the pick, but that's not a regression: the cached-lane path already used the last item, so the old code was already inconsistent there. This PR makes both paths consistent; `gap >= 0` is unaffected.

### Performance

Added multi-lane cases to `tests/bench.bench.ts` (cold-mount `getMeasurements` with variable and uniform sizes, and a `measured`-mode rebuild storm). Mean ms, lower is better:

| scenario | before | after | speedup |
| --- | ---: | ---: | ---: |
| cold mount · variable · lanes=2 · n=100k | 25.86 | 7.52 | 3.4× |
| cold mount · variable · lanes=4 · n=100k | 43.57 | 8.39 | 5.2× |
| cold mount · variable · lanes=8 · n=100k | 80.23 | 8.93 | 9.0× |
| cold mount · uniform · lanes=2 · n=100k | 24.07 | 6.37 | 3.8× |
| cold mount · uniform · lanes=4 · n=100k | 29.83 | 6.67 | 4.5× |
| cold mount · uniform · lanes=8 · n=100k | 52.81 | 8.09 | 6.5× |
| rebuild storm · lanes=2 · ×50 | 180.64 | 27.28 | 6.6× |
| rebuild storm · lanes=4 · ×50 | 320.51 | 41.81 | 7.7× |

Two independent wins: 
1. dropping the per-placement `Map` allocations + sort is a flat constant-factor gain that applies to every case (hence uniform sizes — the common same-height grid — improve too, ~4–6×)
2. removing the backward scan adds a further gain that grows with lane count when heights diverge. Single-lane (`lanes === 1`, the default) is a separate fast path, unchanged within noise.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance of multi-lane virtualization by speeding up lane selection while preserving existing placement geometry and results, including deterministic handling of edge-case ties.
* **Tests**
  * Added new multi-lane differential/oracle coverage to confirm measurement and lane assignment equivalence across fresh builds and incremental rebuilds.
  * Expanded benchmarks for multi-lane scenarios, including cold mounts with uniform/variable lane sizes and rebuild-storm behavior in measurement mode.

### DIFF
--- a/.changeset/blue-icons-appear.md
+++ b/.changeset/blue-icons-appear.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Improve multi-lane virtualization performance: replace the backward scan in getMeasurements with an incremental per-lane argmin (O(lanes) shortest-lane lookup). Placement output is unchanged and the single-lane fast path is untouched.
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -1082,47 +1082,6 @@ export class Virtualizer<
     return this.scrollOffset
   }
 
-  private getFurthestMeasurement = (
-    measurements: Array<VirtualItem>,
-    index: number,
-  ) => {
-    const furthestMeasurementsFound = new Map<number, true>()
-    const furthestMeasurements = new Map<number, VirtualItem>()
-    for (let m = index - 1; m >= 0; m--) {
-      const measurement = measurements[m]!
-
-      if (furthestMeasurementsFound.has(measurement.lane)) {
-        continue
-      }
-
-      const previousFurthestMeasurement = furthestMeasurements.get(
-        measurement.lane,
-      )
-      if (
-        previousFurthestMeasurement == null ||
-        measurement.end > previousFurthestMeasurement.end
-      ) {
-        furthestMeasurements.set(measurement.lane, measurement)
-      } else if (measurement.end < previousFurthestMeasurement.end) {
-        furthestMeasurementsFound.set(measurement.lane, true)
-      }
-
-      if (furthestMeasurementsFound.size === this.options.lanes) {
-        break
-      }
-    }
-
-    return furthestMeasurements.size === this.options.lanes
-      ? Array.from(furthestMeasurements.values()).sort((a, b) => {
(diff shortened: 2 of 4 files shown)


piecyk (approved):
Nice work! I did verify the equivalence claim, the tests also pass against the old implementation at the base commit, and I was able to reproduce the benchmark improvement locally. 👍

PR #1226, opened 2026-07-12
fix(virtual-core): reject 0-size resizeItem when cached measurement exists

### LINKED ISSUE #1188: Dynamic sizes lit demo crashes/ loads very slowly
### Describe the bug

Your demo with the latest versions of `lit-virtual` (3.13.29) and `virtual-core` (3.17.0) either loads very slowly (> 1 minute) or just crashes the browser.

I was really happy a new version was available, because 3.16 (and some versions before) broke the `scrollToIndex` functionality in Lit, but this release might unfortunately be worse.
Once the page actually loads the `scrollToIndex` does seem to work 😁 

https://stackblitz.com/edit/vitejs-vite-wvfdjawc?file=package.json,index.html,tsconfig.json,src%2Fmain.ts&terminal=dev

### Steps to reproduce

I isolated the `RowVirtualizerDynamic` scenario and removed the other 2 that were on the page.
There is no real reproduction path, just try to load the page.

It can take over a minute for the page to load sometimes. And sometimes the browser just crashes.

### Expected behavior

The page loads fast and the scrolling to halfway works.

Often

### Screenshots or Videos

_No response_

### Platform

MacOS 26.5
Chrome 148.0.7778.215
Firefox 148.0.7778.215

### tanstack-virtual version

3.17.0

### TypeScript version

5.6.3

### Additional context

@tanstack/lit-virtual": "3.13.29"

I removed the generation of random paragraphs with `faker` and just have every item be the same string, that didn't solve anything.

### Terms & Code of Conduct

hiro-nikaitou:
## Description

When an element is temporarily hidden (e.g., `display:none` on the container), the ResizeObserver fires with size 0. This overwrites the valid cached measurement in `resizeItem`, causing `calculateRangeImpl` to include all items in the visible range (since all items now have `end === 0`), which results in a browser freeze or `Maximum update depth exceeded` error.

## Fix

Guard `resizeItem` against accepting a 0-size measurement when the item already has a previously cached size. First-time measurements of genuinely 0-height items remain accepted (no prior cache entry).

## Tests

- Verify 0-size is rejected when previous non-zero measurement exists in cache
- Verify first-time 0-size measurement is accepted (no prior cache)
- Verify 0-size → non-zero → 0-size cycle: non-zero measured after first 0 accepted, then 0 rejected after non-zero cached

Fixes #1188

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented temporary zero-size measurements from overwriting previously cached item sizes.
  * Reduced unnecessary re-renders and avoided infinite render loops when virtualized elements are temporarily hidden.
  * Improved range calculations when items briefly report zero dimensions.

### DIFF
--- a/.changeset/zero-size-reject-measurement.md
+++ b/.changeset/zero-size-reject-measurement.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Fix: reject 0-size ResizeObserver measurement when item has cached size
+
+Prevent 0-size measurements from overwriting a previously cached size in resizeItem, avoiding infinite re-render loops when elements are temporarily hidden (e.g., display:none causes ResizeObserver to deliver 0 size, which triggers calculateRangeImpl to include all items and freeze the browser).
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -1515,6 +1515,12 @@ export class Virtualizer<
     }
 
     const itemSize = this.itemSizeCache.get(key) ?? cachedSize
+
+    // Prevent 0-size measurements from overwriting a previously cached size.
+    // This avoids infinite re-renders when elements are temporarily hidden
+    // (e.g., display:none), causing ResizeObserver to deliver 0 size.
+    if (size === 0 && this.itemSizeCache.has(key)) return
+
     const delta = size - itemSize
 
     if (delta !== 0) {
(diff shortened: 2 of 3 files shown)


piecyk (commented):
We already have an opt-in solution for this use case with useCachedMeasurements:

https://tanstack.com/virtual/latest/docs/api/virtualizer#usecachedmeasurements

It was added specifically for lists that are temporarily hidden with display: none, where ResizeObserver reports 0 and would otherwise overwrite the cached measurements.

This PR moves that behavior into resizeItem globally. That means an item with a cached measurement could never legitimately resize to 0, and it would also affect custom measureElement implementations and direct resizeItem calls.

PR #1228, opened 2026-07-14
Angular chat & Angular 20

benjavicente:
## 🎯 Changes

- Bumps Angular to 20, since [version 19 is no longer supported](https://angular.dev/reference/releases#:~:text=Angular%20versions%20v2%20to%20v19%20are%20no%20longer%20supported.). Upgrades TypeScript to be [inside the range Angular supports](https://angular.dev/reference/versions#:~:text=20.2.x%20%7C%7C%2020.3,%5E6.5.3%20%7C%7C%20%5E7.4.0).
- Added a chat example for Angular. It was requested here https://github.com/TanStack/virtual/discussions/1082

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/virtual/blob/main/CONTRIBUTING.md).
- [X] I have tested this code locally with `pnpm run test:pr`.

## 🚀 Release Impact

- [X] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **New Features**
  * Added an Angular chat example demonstrating end-anchored virtual scrolling, history loading, new messages, and streaming replies.
  * Added the chat example to the Angular examples documentation.
* **Breaking Changes**
  * Angular Virtual now requires Angular 20 or newer.
* **Documentation**
  * Added usage guidance and run instructions for the new chat example.
* **Chores**
  * Updated Angular examples and development tooling to Angular 20 and TypeScript 5.9.3.

### DIFF
--- a/.changeset/hot-points-crash.md
+++ b/.changeset/hot-points-crash.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/angular-virtual': major
+---
+
+require angular 20 and up following their lts
--- a/examples/angular/chat/src/app/app.component.ts
+++ b/examples/angular/chat/src/app/app.component.ts
@@ -1,0 +1,230 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  DestroyRef,
+  afterNextRender,
+  afterRenderEffect,
+  inject,
+  signal,
+  viewChild,
+  viewChildren,
+} from '@angular/core'
+import { injectVirtualizer } from '@tanstack/angular-virtual'
+import type { ElementRef } from '@angular/core'
+
+type Message = {
+  id: string
+  author: 'user' | 'assistant'
+  text: string
+}
+
+const replies = [
+  'I can break that into the smallest next step and keep the current viewport pinned while this answer grows.',
+  'Older messages are loaded above the viewport. The visible row keeps the same screen position after the prepend.',
+  'When the thread is not at the bottom, new output waits below without pulling the reader away from history.',
+]
+
+const makeMessage = (index: number): Message => ({
+  id: `message-${index}`,
+  author: index % 4 === 0 ? 'user' : 'assistant',
+  text:
+    index % 4 === 0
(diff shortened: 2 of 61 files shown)


.changeset/hot-points-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/angular-virtual': minor
```
piecyk:
```suggestion
'@tanstack/angular-virtual': major
```

This changeset should be major rather than minor. Changing the @angular/core peer dependency to >=20.0.0 drops support for existing consumers using Angular versions below 20, so this is a breaking change.

piecyk (changes requested):

.changeset/hot-points-crash.md
benjavicente: Done!

examples/angular/chat/src/app/app.component.ts
```diff
@@ -0,0 +1,230 @@
...
+  private nextMessageIndex = initialMessages.length
+  private autoHistoryTimer: number | undefined
+  private streamTimer: number | undefined
+  private autoHistoryEnabled = false
+
+  // Register before the virtualizer so item measurement runs before its
+  // post-render scroll reconciliation, matching React refs before layout effects.
+  private readonly measureItems = afterRenderEffect({
+    mixedReadWrite: () => this.measureVirtualItems(),
+  })
```
benjavicente, line 116: I'm looking if this could be improved. I don't like that if this effect is placed after `injectVirtualizer`, the `followOnAppend` fails. There should be a simple way to solve this but I haven't found one yet.

benjavicente: Using the plain `effect` works, but ideally `afterRenderEffect` should be used.

piecyk (approved):

PR #1232, opened 2026-07-18
fix(virtual-core): read shouldAdjustScrollPositionOnItemSizeChange from options

### LINKED ISSUE #1227: Up-scroll jank from uncompensated above-viewport resizes
### Describe the bug

# Up-scroll jank from uncompensated above-viewport resizes (regressed in virtual-core 3.14), and the `shouldAdjustScrollPositionOnItemSizeChange` escape hatch is not wired in 3.17.x

## Summary

Two related problems in `Virtualizer.resizeItem` affect lists with **dynamically measured** rows (rich text / images / fonts, where measured height ≠ `estimateSize`):

1. Since ~3.14, the default scroll-position compensation on item resize is **skipped while scrolling up**, causing visible jank on up-scroll.
2. The documented workaround — passing `shouldAdjustScrollPositionOnItemSizeChange` — **does nothing in 3.17.x** because the option is never read from `options`, so there is no way to opt out.

## Versions

- `@tanstack/virtual-core`: **last good 3.13.12, first bad 3.16.0** (bisected via `@tanstack/react-virtual` 3.13.12 → 3.14.0). Also present in 3.17.1.
- `@tanstack/react-virtual` 3.14.x, React 19.

## 1) Behavior change in `resizeItem` (3.14+)

**3.13.12 (good)** — compensate whenever the resized item is above the viewport:

```js
item.start < this.getScrollOffset() + this.scrollAdjustments
```

**3.16.0 (bad)** — same, but excluded during backward scroll:

```js
itemStart < this.getScrollOffset() + this.scrollAdjustments && this.scrollDirection !== "backward"
```

**3.17.1** — variant of the same exclusion:

```js
itemStart < this.getScrollOffset() + this.scrollAdjustments
  && (!this.itemSizeCache.has(key) || this.scrollDirection !== "backward")
```

When rows are measured lazily and their real size differs from `estimateSize`, they are corrected as they scroll in from the top. Without compensation during backward scroll, each correction shifts the content below it downward → **stutter while scrolling up** (scrolling down is unaffected, since newly measured rows enter below the viewport). Anchored side content (e.g. connector lines positioned from item offsets) also drifts out of sync.

The in-source comment states this exclusion was added to fix "items jump while scrolling up" for the common case — but for dynamically-measured lists it *causes* exactly that.

## 2) The escape hatch is broken in 3.17.x

The same comment says: *"Users who want the old behavior can pass `shouldAdjustScrollPositionOnItemSizeChange`."*

In 3.17.1, `resizeItem` reads it off the **instance**:

```js
this.shouldAdjustScrollPositionOnItemSizeChange !== void 0
  ? this.shouldAdjustScrollPositionOnItemSizeChange(item, delta, this)
  : /* default with backward exclusion */
```

…but `setOptions` only stores options into `this.options` (`this.options = merged`). `this.shouldAdjustScrollPositionOnItemSizeChange` is declared on the class yet **never assigned**, so it is always `undefined` and the custom predicate is ignored — the default branch always runs. Consequently the option is also absent from the `VirtualizerOptions` input type. There is no working way to restore the pre-3.14 behavior.

## Impact

Up-scroll jank on any virtualized list with dynamic measurement and imperfect estimates, with **no working workaround**, forcing a pin to `virtual-core@3.13.12`.

## Repro

1. Virtualized list with `measureElement` and rows whose real height differs from `estimateSize`.
2. Scroll down (rows get measured), then scroll back up → content stutters as above-viewport rows are corrected without compensation.
3. Pass `shouldAdjustScrollPositionOnItemSizeChange: (item, _d, i) => item.start < i.getScrollOffset() + i.scrollAdjustments` → no effect.

## Suggested fix

- Read the option from `this.options.shouldAdjustScrollPositionOnItemSizeChange` (or assign it onto the instance in `setOptions`) so the escape hatch works again, and expose it on `VirtualizerOptions`.
- Reconsider the `scrollDirection !== "backward"` default: it trades one jank class for another. At minimum a working opt-out is needed for dynamically-measured lists.

https://vitejsvitefd5uzfmy-zaso--5173--87cf54cd.local-credentialless.webcontainer.io/

### Steps to reproduce

1. Open the reproduction (StackBlitz link above). It's a virtualized list using
   `measureElement`, where each row's real height differs from `estimateSize`.
2. Scroll DOWN into the middle of the list so the rows get measured.
3. Click "Grow rows" — this resizes already-measured rows (simulates images/fonts
   or async content settling after the first measure).
4. Scroll UP slowly and watch the list content.
5. On virtual-core 3.14+ (repro ships @tanstack/react-virtual 3.14.6 → virtual-core
   3.17.4) the rows jump/stutter while scrolling up. Set the dependency to
   "@tanstack/react-virtual": "3.13.12" (→ virtual-core 3.13.12) and repeat: the
   list stays anchored, no jank.
6. The repro passes `shouldAdjustScrollPositionOnItemSizeChange` (the documented
   opt-out). Note it has no effect on 3.17.x.

### Expected behavior

When a row above the viewport changes size, the virtualizer should compensate
scrollTop so the visible content stays anchored — regardless of scroll direction,
as it did up to virtual-core 3.13.12. Scrolling up should remain smooth.

Since 3.14 the default skips this compensation while scrolling backward
(`scrollDirection !== "backward"` added to the resizeItem condition), so every
above-viewport correction visibly shifts content during up-scroll. At minimum the
documented `shouldAdjustScrollPositionOnItemSizeChange` option should work so the
pre-3.14 behavior can be restored — but in 3.17.x it is read from the instance
(`this.shouldAdjust…`) while `setOptions` only stores it on `this.options`, so it
is always undefined and ignored.

Always

### Screenshots or Videos

_No response_

### Platform

macOS

### tanstack-virtual version

v3.14.6

### TypeScript version

_No response_

### Additional context

Bisected in @tanstack/virtual-core:
- last good: 3.13.12
- first bad: 3.16.0  (reached via @tanstack/react-virtual 3.14.0; also in 3.17.1/3.17.4)

Root cause — the default scroll-compensation condition in `resizeItem` gained a
backward-scroll exclusion:

  3.13.12:  item.start < this.getScrollOffset() + this.scrollAdjustments
  3.16.0:   ... && this.scrollDirection !== "backward"
  3.17.x:   ... && (!this.itemSizeCache.has(key) || this.scrollDirection !== "backward")

For dynamically measured rows whose size differs from estimateSize (and that can
resize again after their first measure — images/fonts/async content), corrections
to above-viewport rows are no longer compensated while scrolling up, so the visible
content shifts → up-scroll jank. Down-scroll is unaffected (new rows enter below
the viewport).

Broken escape hatch in 3.17.x — `resizeItem` reads
`this.shouldAdjustScrollPositionOnItemSizeChange`, but `setOptions` only stores the
merged options on `this.options` (`this.options = merged`); the instance field is
never assigned, so it is always undefined and the custom predicate is ignored. The
option is therefore also absent from the `VirtualizerOptions` input type. There is
no working way to restore the pre-3.14 behavior.

Suggested fix: read the option from `this.options.shouldAdjustScrollPositionOnItemSizeChange`
(or assign it in setOptions) and expose it on VirtualizerOptions; and/or reconsider
the `scrollDirection !== "backward"` default.

### Terms & Code of Conduct

xianjianlf2:
## 🎯 Changes

Closes #1227.

**Problem:** The documented option `shouldAdjustScrollPositionOnItemSizeChange` is dead code. Passing it to a virtualizer has no effect: `resizeItem` never sees it.

**Root cause:** `resizeItem` reads `this.shouldAdjustScrollPositionOnItemSizeChange` — an instance field that `setOptions` never assigns — and the option is missing from `VirtualizerOptions` entirely. The CHANGELOG (around line 110) documents the option as public API, so this is a wiring bug rather than a design change.

**Fix:** Adds `shouldAdjustScrollPositionOnItemSizeChange` to `VirtualizerOptions`, and makes `resizeItem` read it options-first with a fallback to the instance field for backward compatibility with code that assigned the field directly.

**Testing:** 3 new tests — overriding the default backward-scroll adjustment, the predicate receiving `(item, delta, instance)`, and returning `false` suppressing the default adjustment. 122/122 vitest passing; `tsc`, eslint and prettier clean.

**Scope note:** The issue also asks to revisit the backward-skip default behavior (part 2). This PR deliberately does not touch that — it only wires the documented option so users can opt out today. The default-behavior discussion can proceed separately.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to control whether scroll position adjusts when item sizes change.
  * The callback receives the affected item, size difference, and virtualizer instance for custom behavior.

* **Bug Fixes**
  * Fixed the documented scroll-adjustment option so it now takes effect.
  * Preserved existing behavior when the new option is not provided.

### DIFF
--- a/.changeset/wire-should-adjust-scroll-position-option.md
+++ b/.changeset/wire-should-adjust-scroll-position-option.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Wire `shouldAdjustScrollPositionOnItemSizeChange` into `VirtualizerOptions` so it can actually be supplied. `resizeItem` read the predicate from an instance field that `setOptions` never assigned, so passing the documented escape hatch through options silently did nothing and the default backward-scroll compensation skip could not be opted out of. The predicate is now read from options first, falling back to a directly-assigned instance field for back-compat.
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -369,6 +369,11 @@ export interface VirtualizerOptions<
   useAnimationFrameWithResizeObserver?: boolean
   laneAssignmentMode?: LaneAssignmentMode
   useCachedMeasurements?: boolean
+  shouldAdjustScrollPositionOnItemSizeChange?: (
+    item: VirtualItem,
+    delta: number,
+    instance: Virtualizer<TScrollElement, TItemElement>,
+  ) => boolean
 }
 
 type ScrollState = {
@@ -1547,10 +1552,21 @@ export class Virtualizer<
         this.scrollState?.behavior !== 'smooth' &&
         this.getVirtualDistanceFromEnd() <= this.options.scrollEndThreshold
       const prevTotalSize = wasAtEnd ? this.getTotalSize() : 0
+      // The predicate can be supplied via options (#1227) or, for
+      // back-compat, by assigning it directly on the instance.
+      const shouldAdjustScrollPositionOnItemSizeChange:
+        | ((
+            item: VirtualItem,
+            delta: number,
+            instance: Virtualizer<TScrollElement, TItemElement>,
+          ) => boolean)
+        | undefined =
+        this.options.shouldAdjustScrollPositionOnItemSizeChange ??
+        this.shouldAdjustScrollPositionOnItemSizeChange
       const shouldAdjustScroll =
         this.scrollState?.behavior !== 'smooth' &&
-        (this.shouldAdjustScrollPositionOnItemSizeChange !== undefined
-          ? this.shouldAdjustScrollPositionOnItemSizeChange(
+        (shouldAdjustScrollPositionOnItemSizeChange !== undefined
+          ? shouldAdjustScrollPositionOnItemSizeChange(
               // The callback expects a VirtualItem; build one lazily only
               // when the consumer actually supplied a custom predicate.
               this.measurementsCache[index] ?? {
--- a/packages/virtual-core/tests/index.test.ts
+++ b/packages/virtual-core/tests/index.test.ts
@@ -1,4 +1,5 @@
 import { expect, test, vi } from 'vitest'
+import type { VirtualItem } from '../src/index'
 import {
   Virtualizer,
   _resetIOSDetectionForTests,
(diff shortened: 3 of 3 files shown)


piecyk:
@xianjianlf2 Thanks for the PR and the thorough write-up!

One correction on framing: this isn't dead code. shouldAdjustScrollPositionOnItemSizeChange is by design a property on the virtualizer instance, not a VirtualizerOptions field — it's documented under Virtualizer Instance and is meant to be assigned imperatively

The intent behind making it an instance field (rather than an option) is that it's a rarely-used escape hatch that's convenient to set or clear imperatively from an event handler, without threading it through the options object on every render.

So the reported problem is really that passing it via options doesn't work — which is expected, since it was never an option. I'll grant that the CHANGELOG's "supply" wording is ambiguous and reads like an option, so I understand how this happened.

Given that, I'd lean toward fixing this as a documentation clarification rather than expanding the API surface.

Issue #580, opened 2023-09-08
`getItemKey` is called with invalid index


lesha1201:
### Describe the bug

`getItemKey` might be called with an index which is out of range of the current `count` . It can happen because `ResizeObserver` might be called with the previous nodes and new nodes after the list is changed. For example,

```
[
  <previous-node data-index="0" />,
  <previous-node data-index="1" />,
  <new-node data-index="0" />,
]
```

Previously, we were checking if there is an item with such index in `measurementsCache` but now we don't check it and because of that `getItemKey` is called with a wrong index.

https://github.com/TanStack/virtual/pull/521/files#diff-85d1b85f3d4e501779ca41e51cd38cf7f61a2b7ffddaf958cd7e79c2787c36f1L564-L569

### Expected behavior

Not sure what's the right behavior. I guess we should call `this.observer.unobserve(node)` for the previous nodes here and at the same time to not call `getItemKey` with the previous indexes.

I can accomplish this partially by falling back to an index in `getItemKey`, e.g.:

```ts
const getItemKey = (index) => data[index]?.key || index
```

In this case it will unobserver the previous nodes. But I'm not sure if it's intended way of using `getItemKey` and it doesn't clear `measureElementCache` (`this.measureElementCache.delete(elementKey)`).

### tanstack-virtual version

v3.0.0-beta.59

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

PR #1246, opened 2026-07-30
fix(virtual-core): ignore stale connected measurements

LeoLB-Wang:
## Changes
- Ignore connected measured nodes whose data-index falls outside the current count before any key lookup or measurement callback.
- Add a 22-to-1 count-shrink regression that independently exercises the ref and delayed ResizeObserver ingress paths.
- Add a patch changeset for @tanstack/virtual-core.

## Why
#1148 safely handles disconnected stale nodes. A node can also remain connected while React applies a count shrink; its old data-index can then reach getItemKey, which commonly indexes application data and may throw.

## Testing
- pnpm run test:pr
- pnpm --filter @tanstack/virtual-core test:lib -- index.test.ts

## Related
- #580
- #1147

## Checklist
- [x] Followed the contributing guide.
- [x] Tested locally with pnpm run test:pr.
- [x] Added a changeset for the published package.

## Summary by CodeRabbit

* **Bug Fixes**
  * Virtualized lists now safely ignore measurements from stale items after the item count decreases.
  * Prevents invalid measurement updates and related errors from outdated connected elements.
  * Improves stability when items are removed while their previous elements remain connected or trigger resize events.
  * Ensures only currently valid items are measured and updated.

* **Tests**
  * Added regression coverage for stale measurements and resize events after list size changes.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -489,7 +489,7 @@ export class Virtualizer<
               return
             }
 
-            if !this.isIndexInRange(index) {
+            if (!this.isIndexInRange(index)) {
               return
             }
 
@@ -1174,6 +1174,9 @@ export class Virtualizer<
       key: false,
     },
   )
+
+  private isIndexInRange = (index: number): boolean =>
+    index >= 0 && index < this.options.count
 
   private getMeasurements = memo(
     () => [this.getMeasurementOptions(), this.itemSizeCacheVersion],
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -489,6 +489,10 @@ export class Virtualizer<
               return
             }
 
+            if (index < 0 || index >= this.options.count) {
+              return
+            }
+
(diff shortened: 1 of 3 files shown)


packages/virtual-core/src/index.ts
```diff
@@ -489,6 +489,10 @@ export class Virtualizer<
               return
             }
 
+            if (index < 0 || index >= this.options.count) {
```
piecyk: Let's add private helper like isIndexInRange, then we can
```suggestion
            if !this.isIndexInRange(index) {
```

Like 

```tsx
   * Whether an index still addresses a live item. A measured node carries its
   * index in a DOM attribute, so the index can outlive the item it referred to
   * when `count` shrinks — and `getItemKey` commonly indexes application data,
   * which then throws. Callers must check this before deriving a key from an
   * index that came out of the DOM.
   */
  private isIndexInRange = (index: number): boolean =>
    index >= 0 && index < this.options.count
```

piecyk:
Thanks @LeoLB-Wang

PR #1253, opened 2026-08-12
fix(lit-virtual): add setOptions to VirtualizerController for reactive option updates

### LINKED ISSUE #1251: lit-virtual: VirtualizerController never applies updated options after construction
### Describe the bug

`VirtualizerController` (`packages/lit-virtual/src/index.ts`) never calls `setOptions()` after the initial construction, so reactive option updates (a changed `count`, `estimateSize`, `getScrollElement`, etc.) never reach the underlying `Virtualizer` instance.

```ts
constructor(
  host: ReactiveControllerHost,
) {
  ...
  this.virtualizer = new Virtualizer(resolvedOptions) // options captured once, here
  ;(this.host = host).addController(this)
}

hostUpdated() {
  this.virtualizer._willUpdate() // no setOptions() call
}
```

Every other adapter calls `setOptions()` on every render/update before `_willUpdate()`. For comparison, `react-virtual`'s `useVirtualizer`:

```ts
...
return instance._willUpdate()
```

and `angular-virtual`:

```ts
virtualizer.setOptions(resolvedOptions())
...
reactiveVirtualizer()._willUpdate()
```

`lit-table`'s equivalent `TableController` (same Lit `ReactiveController` shape, different package) does call the table-core equivalent of `setOptions` on every access, so this isn't a Lit-specific limitation, `lit-virtual` just doesn't do it.

Since `VirtualizerController` is meant to be instantiated once in the host's constructor (the documented usage pattern, same as any other Lit `ReactiveController`), there's no way for a consumer to get a changed `count` or other option applied short of reaching into `getVirtualizer()` and calling `.setOptions()` manually, which isn't documented anywhere and defeats the point of passing options to the constructor in the first place.

```ts
import { LitElement, html } from 'lit'
import { VirtualizerController } from '@tanstack/lit-virtual'

class MyList extends LitElement {
  count = 10

  virtualizerController = new VirtualizerController(this, {
    count: this.count,
    estimateSize: () => 35,
    getScrollElement: () => this.renderRoot.querySelector('#scroller'),
  })

  grow() {
    this.count = 1000
    this.requestUpdate()
  }

  render() {
    const virtualizer = this.virtualizerController.getVirtualizer()
    // virtualizer.options.count is still 10 after grow(), even though
    // this.count is 1000 and the component re-rendered
    return html`<div id="scroller">${virtualizer.getVirtualItems().length} items</div>`
  }
}
```

### Steps to reproduce

1. Create a `VirtualizerController` with an initial `count`.
2. Change the value backing `count` and call `requestUpdate()` on the host.
3. Read `virtualizer.options.count` (or observe the rendered item count), it still reflects the value from construction time.

### Expected behavior

Option changes on subsequent renders should reach the virtualizer, the same as every other framework adapter.

Worth noting this needs more than adding a `setOptions()` call inside `hostUpdated()`: the constructor is the only place `options` is ever received (`options: VirtualizerOptions<...>`, not a getter), so even a `hostUpdated()` call would just re-apply the same closed-over object from construction time. The API shape itself doesn't have a channel for fresh options to flow in after construction. `lit-table`'s `TableController` sidesteps this by exposing a `.table(tableOptions, selector)` method that's called fresh from `render()` on every pass rather than once from the constructor, that shape (or an equivalent public `setOptions()` method consumers call from `render()`) looks like the fix, not just wiring `hostUpdated()` to an already-stale value.

### Platform

n/a, framework-adapter logic (`packages/lit-virtual/src/index.ts`)

waterWang:
## Description

Fixes #1251

 () never calls  after the initial construction, so reactive option updates (a changed , , , etc.) never reach the underlying  instance.

### Root cause

The constructor captures options once and passes them to . The  lifecycle method only calls  without re-applying options. Every other framework adapter (, ) calls  before every .

### Fix

1. **Added  method** to  — stores the options on the instance, merges with new partial options, and calls  + . Matches the pattern used by 's  and 's injectable.

2. **Stored  as a private field** — enables the  method to merge partial updates with the base options (including defaults from  and  subclass constructors).

### Changes

- : Added  method, stored  as private field
- : Added test verifying  renders items beyond the initial count of 10

## Summary by CodeRabbit

* **New Features**
  * Added the ability to update virtualizer options dynamically during rendering.
  * Changes to options now immediately refresh the displayed item range and count.

* **Tests**
  * Added coverage confirming that updated item counts render newly available items correctly.

### DIFF
--- a/packages/lit-virtual/src/index.ts
+++ b/packages/lit-virtual/src/index.ts
@@ -18,12 +18,15 @@ class VirtualizerControllerBase<
 
   private readonly virtualizer: Virtualizer<TScrollElement, TItemElement>
 
+  private options: VirtualizerOptions<TScrollElement, TItemElement>
+
   private cleanup: () => void = () => {}
 
   constructor(
     host: ReactiveControllerHost,
     options: VirtualizerOptions<TScrollElement, TItemElement>,
   ) {
+    this.options = options
     const resolvedOptions: VirtualizerOptions<TScrollElement, TItemElement> = {
       ...options,
       onChange: (instance, sync) => {
@@ -37,6 +40,21 @@ class VirtualizerControllerBase<
 
   public getVirtualizer() {
     return this.virtualizer
+  }
+
+  public setOptions(
+    options: Partial<VirtualizerOptions<TScrollElement, TItemElement>>,
+  ) {
+    this.options = { ...this.options, ...options }
+    const resolvedOptions: VirtualizerOptions<TScrollElement, TItemElement> = {
+      ...this.options,
+      onChange: (instance, sync) => {
+        this.host.updateComplete.then(() => this.host.requestUpdate())
+        this.options.onChange?.(instance, sync)
+      },
+    }
+    this.virtualizer.setOptions(resolvedOptions)
+    this.virtualizer._willUpdate()
   }
 
   hostConnected() {
--- a/packages/lit-virtual/tests/index.test.ts
+++ b/packages/lit-virtual/tests/index.test.ts
@@ -119,7 +119,10 @@ test('should apply updated options via setOptions', async () => {
   class ListSetOptions extends LitElement {
     private scrollElementRef: Ref<HTMLDivElement> = createRef()
 
-    private virtualizerController: VirtualizerController<HTMLDivElement, Element>
+    private virtualizerController: VirtualizerController<
+      HTMLDivElement,
+      Element
+    >
 
     constructor() {
       super()
@@ -169,14 +172,23 @@ test('should apply updated options via setOptions', async () => {
           </div>
         </div>
         <style>
-          .list { border: 1px solid #e6e4dc; max-width: 100%; }
-          .scroll-container { height: ${height}px; width: ${width}px; overflow-y: auto; }
+          .list {
+            border: 1px solid #e6e4dc;
+            max-width: 100%;
+          }
+          .scroll-container {
+            height: ${height}px;
+            width: ${width}px;
+            overflow-y: auto;
+          }
         </style>
       `
     }
   }
 
-  const el = await fixture(html`<test-list-setoptions></test-list-setoptions>` as any)
+  const el = await fixture(
+    html`<test-list-setoptions></test-list-setoptions>` as any,
+  )
   await elementUpdated(el)
   await waitUntil(
     () => el.shadowRoot.querySelector('[data-index="15"]'),
--- a/packages/lit-virtual/tests/index.test.ts
+++ b/packages/lit-virtual/tests/index.test.ts
@@ -113,3 +113,74 @@ test('should render virtual items', async () => {
     'Element did not render virtual items',
   )
 })
+
+test('should apply updated options via setOptions', async () => {
+  @customElement('test-list-setoptions' as any)
+  class ListSetOptions extends LitElement {
+    private scrollElementRef: Ref<HTMLDivElement> = createRef()
+
+    private virtualizerController: VirtualizerController<HTMLDivElement, Element>
+
+    constructor() {
+      super()
+      this.virtualizerController = new VirtualizerController(this, {
+        getScrollElement: () => this.scrollElementRef.value,
+        count: 10,
+        estimateSize: () => 50,
+        observeElementRect: (_, cb) => {
+          cb({ height, width })
+        },
+      })
+    }
+
+    render() {
+      this.virtualizerController.setOptions({
+        count: 20,
+      })
+      const virtualizer = this.virtualizerController.getVirtualizer()
+      const virtualRows = virtualizer.getVirtualItems()
+      return html`
+        <div class="list scroll-container" ${ref(this.scrollElementRef)}>
+          <div
+            style="position: relative; height: ${virtualizer.getTotalSize()}px; width: 100%;"
+          >
+            <div
+              style="position:absolute;top:0;left:0;width:100%;transform:translateY(${virtualRows[0]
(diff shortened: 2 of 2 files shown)


MILLERMARRU (commented):
This closes an obvious API gap, every other framework adapter (React, Vue, Solid) can update virtualizer options reactively, but lit-virtual's `VirtualizerControllerBase` only accepted options at construction time. The implementation stores `this.options` and merges partial updates on `setOptions`, re-wrapping `onChange` the same way the constructor does, which keeps the host's `requestUpdate` callback wired correctly on every option change. Calling `this.virtualizer._willUpdate()` after `setOptions` looks necessary and correct, skipping it would leave stale virtual items rendered until the next unrelated re-render. The added test exercises this by calling `setOptions` inside `render()`, which is the realistic Lit usage pattern.

packages/lit-virtual/src/index.ts
```diff
@@ -39,6 +42,21 @@ class VirtualizerControllerBase<
     return this.virtualizer
   }
 
+  public setOptions(
+    options: Partial<VirtualizerOptions<TScrollElement, TItemElement>>,
+  ) {
+    this.options = { ...this.options, ...options }
```
piecyk, line 48: A plain spread lets an explicit `undefined` wipe a stored option: `setOptions({ estimateSize: undefined })` leaves `estimateSize` undefined for good, and core then falls back to its own default rather than the value set earlier.

Core's `setOptions` guards against exactly this (it skips keys whose value is `undefined`) — worth matching here.

packages/lit-virtual/src/index.ts
```diff
@@ -39,6 +42,21 @@ class VirtualizerControllerBase<
...
+    const resolvedOptions: VirtualizerOptions<TScrollElement, TItemElement> = {
```
piecyk, line 49: This `resolvedOptions` block is a copy of the one in the constructor, and the two close over different things (`options` vs `this.options`). Harmless today, but a shared private `resolveOptions(options)` used by both would stop them drifting.

packages/lit-virtual/tests/index.test.ts
```diff
@@ -113,3 +113,86 @@ test('should render virtual items', async () => {
...
+        count: 10,
+        estimateSize: () => 50,
+        observeElementRect: (_, cb) => {
+          cb({ height, width })
+        },
+      })
+    }
+
+    render() {
+      this.virtualizerController.setOptions({
```
piecyk, line 140: Calling `setOptions` inside `render()` means the count is 20 from the very first paint, so nothing is actually updated reactively — the test would pass without any reactivity at all.

Driving it from a reactive `@property` in `willUpdate` is what the feature is for, and it avoids modelling a pattern that mutates controller state during render.

packages/lit-virtual/tests/index.test.ts
```diff
@@ -113,3 +113,86 @@ test('should render virtual items', async () => {
...
+      `
+    }
+  }
+
+  const el = await fixture(
+    html`<test-list-setoptions></test-list-setoptions>` as any,
+  )
+  await elementUpdated(el)
+  await waitUntil(
+    () => el.shadowRoot.querySelector('[data-index="15"]'),
```
piecyk, line 194: This is what fails. Index 15 is never rendered: the viewport is 400px with `estimateSize: () => 50`, so only about indexes 0–9 render at scroll offset 0 — with `count: 10` and with `count: 20` alike. Raising the count doesn't widen the rendered window, so the test would fail even if `setOptions` were perfect.

`getTotalSize()` is what actually moves (500 → 1000). Assert that instead, and assert the initial 500 too, so the test also covers the constructor options — right now nothing checks the starting count of 10 took effect.

The follow-up `expect(...).toBeTruthy()` on line 197 is also redundant; `waitUntil` already guarantees it.

piecyk (changes requested):
Thanks for this — the implementation itself is right. I checked it separately: driving `setOptions({ count })` from a reactive property moves `getTotalSize()` from 500 to 1000 and re-renders the host.

The blocker is the new test, which fails on this branch:

```
× should apply updated options via setOptions 1013ms
  Error: Timeout: Element did not render items beyond initial count of 10
```

It isn't flaky — the assertion can never pass. Details inline.

Two other things before merge:

**Changeset.** There isn't one. It should be `minor` rather than `patch`, since this adds public API (the repo maps `feat` → minor, `fix` → patch). Worth retitling the PR to `feat(lit-virtual):` to match. Add `.changeset/lucky-pugs-shave.md`:

````markdown
---
'@tanstack/lit-virtual': minor
---

Add `setOptions` to `VirtualizerController` and `WindowVirtualizerController` so virtualizer options can be updated reactively after construction.

Previously the controller only accepted options once, in its constructor, so options derived from reactive Lit properties (`count`, `estimateSize`, `horizontal`, …) could never be re-applied — the only escape hatch was `getVirtualizer().setOptions(...)`, which drops the controller's `onChange` wrapper and stops the host from re-rendering.

`setOptions` takes a partial set of options, merges them over the options the controller already holds, re-applies the host `requestUpdate` wrapper around `onChange`, and flushes the change so the host re-renders:

```ts
willUpdate(changed: PropertyValues<this>) {
  if (changed.has('count')) {
    this.virtualizerController.setOptions({ count: this.count })
  }
}
```
````

**Docs.** `docs/framework/lit/lit-virtual.md` doesn't mention `setOptions`. A short section would help, especially to note that the controller merges partial options while `virtualizer.setOptions` replaces them wholesale — mixing the two leaves the controller's cached `this.options` stale.

PR #1254, opened 2026-08-12
fix(virtual-core): add touch provenance flag to iOS deferral gate

### LINKED ISSUE #1250: iOS scroll-adjustment deferral engages for programmatic scrolls, making scrollToIndex landings paint sagged then snap
### Describe the bug

On iOS-family WebKit, the iOS scroll-adjustment deferral engages for **programmatic** scrolls, not just touch-driven ones. The result is that dynamic-measurement compensation which would normally apply pre-paint is deferred past a paint, so a `scrollToIndex`/`scrollToOffset` landing paints at a visibly wrong offset and then snaps into place a beat later.

The deferral gate is `isScrolling`, and `isScrolling` carries no provenance — `observeOffset` sets it from *any* `scroll` event, including the ones a programmatic scroll write generates itself:

https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts — verified in `3.17.7`:

```js
// observeOffset — any scroll event, no user/programmatic distinction
const createHandler = (isScrolling) => () => {
  offset = readOffset(element)
  fallback?.()
  cb(offset, isScrolling)   // handler = createHandler(true)
}
element.addEventListener('scroll', handler, addEventListenerOptions)
```

```js
// applyScrollAdjustment
if (isIOSWebKit() && (this.isScrolling || this._iosTouching || this._iosJustTouchEnded)) {
  this._iosDeferredAdjustment += delta
  return false
}
```

So the sequence for a programmatic landing is:

1. App calls `scrollToIndex(deepIndex)` → `scrollTop` write → `scroll` event → `isScrolling = true`.
2. The newly mounted rows measure. Any row whose real height differs from `estimateSize` produces an above-fold delta → `applyScrollAdjustment(delta)`.
3. Because `isScrolling` is true, every one of those adjustments is **deferred** instead of applied. Off iOS (and on iOS before this code path existed) they are applied synchronously inside the ResizeObserver callback, i.e. **pre-paint**.
4. The frame paints uncompensated — the list sits offset by the accumulated delta.
5. `isScrolling` flips false ~`isScrollingResetDelay` later → `_flushIosDeferredIfReady()` applies the whole backlog in one write → **visible snap**.

The deferral exists to protect touch momentum (#884 — writing `scrollTop` mid-fling cancels it), which is completely reasonable. But a programmatic scroll has no momentum to protect, so on that path the deferral has no upside and produces a user-visible artifact. Touch provenance is already tracked (`_iosTouching`, `_iosJustTouchEnded`); it is the bare `isScrolling` term that over-captures.

I could not provide a hosted sandbox link that demonstrates this: the code path is gated behind `isIOSWebKit()`, so it only runs when the browser reports an iPhone/iPad UA (or touch-capable MacIntel). A CodeSandbox/StackBlitz preview opened on a desktop UA silently takes the non-iOS branch and looks correct.

The repro below is self-contained and dependency-free beyond `react` + `@tanstack/react-virtual` — drop it into any of the official examples and open it under the conditions in "Steps to reproduce".

```tsx
import { useVirtualizer } from '@tanstack/react-virtual'
import { useEffect, useRef } from 'react'

const COUNT = 500
const LANDING_INDEX = 300
const ESTIMATE = 50
// Every row renders TALLER than estimateSize, so each measurement above the
// landing offset produces a real scroll-anchoring correction. (In our app this
// delta is not synthetic — WebKit's line-box rounding makes the same rows
// measure ~1px taller than they do in Chromium, which is how we hit it.)
const ACTUAL = 58

export default function App() {
  const parentRef = useRef<HTMLDivElement>(null)

    count: COUNT,
    estimateSize: () => ESTIMATE,
  })

  // The "landing": one programmatic jump to a deep index on mount.
  useEffect(() => {
    virtualizer.scrollToIndex(LANDING_INDEX, { align: 'start' })
  }, [])

  return (
    <div ref={parentRef} style={{ height: 600, overflowY: 'auto' }}>
      <div
        style={{
          height: virtualizer.getTotalSize(),
          position: 'relative',
          width: '100%',
        }}
      >
        {virtualizer.getVirtualItems().map((item) => (
          <div
            key={item.key}
            data-index={item.index}
            ref={virtualizer.measureElement}
            style={{
              position: 'absolute',
              top: 0,
              left: 0,
              width: '100%',
              height: ACTUAL, // ← taller than estimateSize
              transform: `translateY(${item.start}px)`,
              borderBottom: '1px solid #ddd',
            }}
          >
            Row {item.index}
          </div>
        ))}
      </div>
    </div>
  )
}
```

### Steps to reproduce

1. Serve the example above and open it in **iOS Safari** — either a real iPhone/iPad, or desktop Safari's Responsive Design Mode with an iPhone selected (RDM spoofs the UA, which is what satisfies `isIOSWebKit()`).
2. Load the page and watch row 300 as it lands. **Do not touch the screen** — the whole point is that this is a purely programmatic scroll.
3. Observe: the list first paints with row 300 offset from where it belongs, then snaps to the correct position shortly after (roughly `isScrollingResetDelay` later).
4. For contrast, open the same page in Chrome, or in Safari on a desktop UA: the landing is correct on the first painted frame, because the adjustments are applied synchronously in the ResizeObserver callback rather than deferred.

### Expected behavior

A programmatic scroll should land compensated on its first painted frame on iOS, exactly as it does on other engines — no sag-then-snap. Deferral should apply to scrolls where a real touch is involved (where cancelling momentum is the concern it was introduced for), not to scrolls the application initiated itself.

Every time

### Screenshots or Videos

Measured in our app with a per-animation-frame sampler recording the landing anchor's viewport offset (WebKit, iPhone UA, 390×874, deferral engaged):

| | anchor transient vs settled | sag frame painted? |
|---|---|---|
| deferral engaged | 4.0px | yes — lands 8.0 → 12.8 → settles 8.8 |
| deferral not engaged | 0.8px | no — 8.0 → 8.8 directly |

That was with a deliberately synthesized measurement delta to make it deterministic. The original user-reported artifact in the real app was a ~10 CSS px whole-list snap, which is very visible.

### Platform

- OS: macOS 26.5.2 (host) — reproduced on iOS Safari (real device) and Safari Responsive Design Mode; also reproduced deterministically in Playwright WebKit 26.5 with an iPhone 13 device profile
- Browser: Safari / WebKit
- Chromium and Firefox are unaffected (the code path is iOS-gated)

### tanstack-virtual version

`@tanstack/react-virtual` v3.14.9 (`@tanstack/virtual-core` v3.17.6). Re-verified the code paths above against `virtual-core` v3.17.7 — unchanged. Bumping 3.17.6 → 3.17.7 does not change the behavior.

### TypeScript version

v6.0.3

### Additional context

**On a possible fix.** The obvious narrowing — dropping the bare `isScrolling` term and gating on `_iosTouching || _iosJustTouchEnded` — is not sufficient on its own: momentum continues well past `touchend`, while `_iosJustTouchEnded` is only true for a 150ms window, so `isScrolling` is presumably in the gate precisely to cover the rest of the fling. Removing it would reintroduce #884.

What seems to be missing is *touch provenance for the whole scroll sequence* rather than a short timer: set a flag on `touchstart`, keep it set for as long as `isScrolling` remains true, and clear it when the scroll fully settles. Deferring on that flag would cover the entire momentum phase (fixing #884 as designed) while leaving app-initiated scrolls — which cannot have momentum to cancel — on the synchronous path. Happy to open a PR along those lines if that direction seems right.

**Relationship to #1189.** I found the draft PR "feat(virtual-core): iOS momentum-safe scroll adjustments via CSS offset", which replaces deferral with a CSS `marginTop` compensation. If that lands, it plausibly removes this symptom as a side effect, since the compensation would be visible immediately rather than withheld until flush. It does list "force-flush CSS offset before programmatic scroll operations", which addresses the inverse direction (a backlog existing *before* a programmatic scroll starts) rather than adjustments arriving *during* one — so I wanted to report this case explicitly in case it is not already covered. It has been a draft since June, so I did not want to assume it was the intended path here.

**Workaround, for anyone hitting this.** We keep the deferral disengaged for a bounded window after our own landing writes: a `requestAnimationFrame` tick clears `isScrolling` on the instance and flushes any backlog, stopping permanently as soon as a real touch is observed. rAF runs before ResizeObserver callbacks within a frame, so measurements arriving later in the same frame compensate pre-paint again. It works, but it pokes at private fields (`_iosDeferredAdjustment`, `_flushIosDeferredIfReady`, `isScrolling`), which is exactly why a supported behavior upstream would be better.

One caveat worth flagging for anyone writing a similar workaround: `isScrolling` is not only the deferral gate — the React adapter also uses it to choose `flushSync` over an async re-render on notify. Clearing it on engines where the deferral does not even engage makes rows lag a frame behind real scrolls. We ended up gating our workaround on a narrower environment predicate than `isIOSWebKit()` itself, because Chromium's mobile emulation on a macOS host reports `platform === 'MacIntel'` with `maxTouchPoints > 0` and therefore satisfies `isIOSWebKit()` while running Blink.

waterWang:
## Problem

On iOS WebKit, the iOS scroll-adjustment deferral engages for **programmatic** scrolls, not just touch-driven ones. The deferral gate is `isScrolling`, and `isScrolling` carries no provenance — `observeOffset` sets it from *any* `scroll` event, including the ones a programmatic scroll write generates itself.

The result is that dynamic-measurement compensation which would normally apply pre-paint is deferred past a paint, so a `scrollToIndex`/`scrollToOffset` landing paints at a visibly wrong offset and then snaps into place a beat later.

## Root Cause

The gate in `applyScrollAdjustment`:

```ts
if (isIOSWebKit() && (this.isScrolling || this._iosTouching || this._iosJustTouchEnded)) {
```

The `this.isScrolling` term over-captures: it's true for ALL scroll events, including the echo of the app's own `scrollTop` write. The existing touch-provenance flags (`_iosTouching`, `_iosJustTouchEnded`) are already correct — they only fire for real touch events — but they only cover the active-touch and 150ms-post-touchend windows. Momentum continues well past that, and `isScrolling` is the only flag that stays true through the rest of the fling.

## Fix

Introduce `_isUserScrolling` — a touch-provenance flag that covers the entire user-initiated scroll sequence:

- **`touchstart`**: set `_isUserScrolling = true`
- **`isScrolling` transitions to false**: clear `_isUserScrolling` (scroll fully settled)
- **Deferral gates**: use `_isUserScrolling` instead of bare `isScrolling`

Touch-initiated sequences defer exactly as before (fixing #884). Programmatic scrolls never set `_isUserScrolling`, so their adjustments stay on the synchronous pre-paint path — landing compensated on the first painted frame.

Fixes #1250

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scrolling behavior on iOS by ensuring deferred adjustments occur only during user-initiated touch scrolling.
  * Prevented programmatic scroll actions from being delayed or incorrectly synchronized.
  * Improved anchor positioning during active touch-based scrolling.

### DIFF
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -425,6 +425,18 @@ export class Virtualizer<
   // iOS momentum scroll (writing scrollTop mid-momentum cancels it).
   // Flushed in a single scrollTo when iOS is fully settled.
   private _iosDeferredAdjustment = 0
+  // Touch provenance for the whole scroll sequence. `isScrolling` is set by
+  // *any* scroll event, including the echo of a programmatic `scrollTop`
+  // write, so it over-captures: on iOS it would defer scroll adjustments
+  // during app-initiated scrolls too, painting a sagged landing then snapping
+  // (#1250). We instead track whether the current scroll sequence began with
+  // a real touch (`touchstart`), keep it set for as long as `isScrolling`
+  // remains true (covering the whole momentum phase), and clear it when the
+  // scroll fully settles. Programmatic scrolls never set it, so their
+  // adjustments stay on the synchronous pre-paint path. Introduced as a
+  // supporting signal for the deferral gates below; touch-driven sequences
+  // defer exactly as before.
+  private _isUserScrolling = false
   // Touch state. iOS WebKit cancels momentum when scrollTop is written, so
   // we defer adjustments not only during `isScrolling` but also through the
   // touchstart→touchend window (active drag) and a short tail after
@@ -697,7 +709,7 @@ export class Virtualizer<
 
     if (
       isIOSWebKit() &&
-      (this.isScrolling || this._iosTouching || this._iosJustTouchEnded)
+      (this._isUserScrolling || this._iosTouching || this._iosJustTouchEnded)
     ) {
       this._iosDeferredAdjustment += delta
       return false
@@ -774,6 +786,7 @@ export class Virtualizer<
     // pending reset of the flag), deferring every adjustment on the new
     // element until its next touch cycle.
     this._iosDeferredAdjustment = 0
+    this._isUserScrolling = false
     this._iosTouching = false
     this._iosJustTouchEnded = false
     this.scrollElement = null
@@ -865,6 +878,13 @@ export class Virtualizer<
             : null
           this.scrollOffset = offset
           this.isScrolling = isScrolling
+          // When the scroll fully settles, touch provenance for the sequence
+          // ends: programmatic scrolls never set it, and a user sequence is
+          // done once the momentum/subsequent scroll events stop. Clearing it
+          // here lets a later app-initiated scroll (which fires its own
+          // scroll events) take the synchronous adjustment path as intended
+          // (#1250).
+          if (!isScrolling) this._isUserScrolling = false
 
           // Flush deferred iOS adjustments if we're now fully settled.
           // "Fully settled" means: not actively scrolling, no finger on
@@ -885,6 +905,7 @@ export class Virtualizer<
       if ('addEventListener' in this.scrollElement) {
         const scrollEl = this.scrollElement as unknown as EventTarget
         const onTouchStart = () => {
+          this._isUserScrolling = true
           this._iosTouching = true
           this._iosJustTouchEnded = false
           if (this._iosTouchEndTimerId !== null && this.targetWindow != null) {
@@ -952,7 +973,7 @@ export class Virtualizer<
         // _flushIosDeferredIfReady handle it once the scroll settles.
         if (
           isIOSWebKit() &&
-          (this.isScrolling || this._iosTouching || this._iosJustTouchEnded)
+          (this._isUserScrolling || this._iosTouching || this._iosJustTouchEnded)
         ) {
           if (anchorDelta !== 0) {
             this._iosDeferredAdjustment += anchorDelta


piecyk:
@waterWang Thanks for the writeup — the diagnosis is right. `isScrolling` genuinely does over-capture, and since `_iosTouching`/`_iosJustTouchEnded` only cover the drag plus a 150 ms tail, removing `isScrolling` from the gate does need something that spans momentum. Two findings before this can land, though.

### 1. `_isUserScrolling` never gets cleared if the touch doesn't scroll

It's set on `touchstart` and cleared in only two places: `cleanup()`, and the `!isScrolling` branch of the scroll callback. That second clear needs a scroll event to have happened — with the default `useScrollendEvent: false`, `!isScrolling` comes from the debounce in `observeOffset`, which is only armed from inside the `scroll` handler.

So any touch that doesn't scroll latches the flag `true` for good: tapping a row, a long-press, a horizontal swipe in a vertical list. Verified at `585b1c7` — `touchstart` + `touchend`, no scroll events, after the grace window:

```
touching: false  justEnded: false  userScrolling: true
applied sync: false  deferred: 20
```

Two consequences after one stray tap: **#1250 still reproduces** (the gate is stuck, so the next `scrollToIndex` defers, paints sagged, snaps), and the deferred delta is *stranded* — `_flushIosDeferredIfReady` only runs from the scroll callback and the touchend timer, and its readiness check (`!isScrolling && !_iosTouching && !_iosJustTouchEnded`) is already satisfied, so nothing re-invokes it. Worth keeping as a rule: the deferral gate and the flush predicate have to agree, otherwise a delta can be deferred by a condition the flush path can never observe.

### 2. A tap-triggered landing still defers — in both this version and mine

Built the issue's repro as a test (500 rows, estimate 50 / actual 58, `scrollToIndex(300)`). The filed case — no touch at all — is fixed. But when the landing is triggered *by* a tap, e.g. tapping a search result whose click handler calls `scrollToIndex`, the call lands inside the post-`touchend` window and the compensation defers anyway:

```
A no-touch  -> deferred: 0   sync writes: 1
B tap       -> deferred: 8   sync writes: 0
```

Since the issue's expectation is the general one ("a programmatic scroll should land compensated on its first painted frame"), closing this needs one more thing: `scrollToOffset`/`scrollToIndex` should also close the post-touchend window, right where they already drop `_iosDeferredAdjustment`. The write those commands are about to make cancels any in-flight momentum by itself, so the window has nothing left to protect. (`_iosTouching` should stay — with a finger still down the user owns the scroll.)

### Suggested shape

Rather than a new field, re-arm the window that already exists. Momentum fires a scroll event per frame, so pushing `_iosJustTouchEnded`'s timer out on each one spans the whole fling and self-terminates 150 ms after the last frame:

```ts
private _armIosTouchWindow = () => {
  if (!isIOSWebKit() || this.targetWindow == null) return
  this._iosJustTouchEnded = true
  if (this._iosTouchEndTimerId !== null) {
    this.targetWindow.clearTimeout(this._iosTouchEndTimerId)
  }
  this._iosTouchEndTimerId = this.targetWindow.setTimeout(() => {
    this._iosJustTouchEnded = false
    this._iosTouchEndTimerId = null
    this._flushIosDeferredIfReady()
  }, 150)
}
```

`onTouchEnd` calls it, and the scroll callback calls it when `isScrolling && this._iosJustTouchEnded`. The gate goes back to `_iosTouching || _iosJustTouchEnded`, matches the flush predicate again, and every piece of state is timer-bounded so nothing can latch. One caveat to be aware of: a self-write echo arriving inside the window also re-arms it, which is part of why fix #2 above is needed rather than optional.

### Also

- **The existing tests need updating with this change.** At `585b1c7`, 10 iOS tests in `packages/virtual-core/tests/index.test.ts` fail — they simulate a live scroll with no touch and assert deferral, which is precisely the behavior #1250 calls a bug. They have to move to touch provenance, and one (`iOS Phase 1: scroll-event after touchend timer cleanup also flushes`) is built entirely on the old premise. Doesn't look like the test job ran on this PR — only CodeRabbit and Socket reported.
- **`touchcancel` isn't handled** (pre-existing): iOS fires it instead of `touchend` when a system gesture steals the touch, which strands `_iosTouching` with no timer to recover it.
- **Naming:** the neighbouring fields are all `_ios*`, and this one is only read on iOS paths.

I have the suggested shape implemented locally with the tests updated and the two new regression cases (tap doesn't latch the gate; momentum re-arms the window) — happy to push it as a commit here or a separate PR, whichever you prefer.

piecyk:
Closing as superseded by #1280, which takes the shape from the review above: the gate is back to `_iosTouching || _iosJustTouchEnded`, momentum scroll events re-arm the existing post-touchend timer so it spans the fling and self-terminates, and `scrollToOffset` / `scrollToIndex` close the window. Both cases found here (a tap that never scrolls latching the gate, and a tap-triggered landing still deferring) are regression tests there, along with the rewritten iOS suite.

Thanks @waterWang for the diagnosis — it was right, and #1280 credits it.

PR #1256, opened 2026-08-18
fix(virtual-core): cancel the isScrolling debounce on scroll-observer cleanup

### LINKED ISSUE #1255: Virtualizer.cleanup() does not clear the isScrolling debounce timer, so an unmounted virtualizer still fires one onChange
Reproduced against `@tanstack/virtual-core@3.17.7` (latest at time of writing, published 2026-07-28).

## The bug

`observeElementOffset` arms a debounce on every `scroll` event to reset `isScrolling` back to `false` after `isScrollingResetDelay` (default 150 ms). `Virtualizer.cleanup()` removes the scroll listener, disconnects the `ResizeObserver` and cancels its `requestAnimationFrame` — but it never clears that debounce timer.

`dist/esm/index.js`, `cleanup()`:

```js
this.cleanup = () => {
  this.unsubs.filter(Boolean).forEach((d) => d());
  this.unsubs = [];
  this.observer.disconnect();
  if (this.rafId != null && this.targetWindow) {
    this.targetWindow.cancelAnimationFrame(this.rafId);
    this.rafId = null;
  }
  // … nothing clears the observeOffset debounce
};
```

and `dist/esm/utils.js`, where the handle lives:

```js
targetWindow.clearTimeout(timeoutId);
timeoutId = targetWindow.setTimeout(() => fn.apply(this, args), ms);
```

`timeoutId` is local to the `debounce` closure, so **no consumer can clear it** — the fix has to be inside the library rather than something a caller can do.

## Why it is reachable by default

This is the default path, not an edge case:

- `useScrollendEvent` defaults to `false` (`dist/esm/index.js`, the default options), so the debounce is used even in browsers that support `scrollend`.
- `supportsScrollend` is `"onscrollend" in window`, and jsdom has no `onscrollend` at all, so every jsdom-based test suite takes this path unconditionally.

## Effect

Unmount a virtualized list inside the 150 ms window after a scroll and one late `(offset, false)` callback still arrives. That runs `maybeNotify()` → `notify()` → the consumer's `onChange` → in React, a `useReducer` dispatch into a tree that no longer exists.

In a browser that dispatch is a silent no-op. Under vitest/jsdom it is not — it surfaces as an update-after-unmount and can fail a test suite that is otherwise correct.

## Suggested fix

Have `observeOffset` return a disposer that clears the pending timeout (or hold the handle on the instance) and call it from `cleanup()` alongside the existing `cancelAnimationFrame`.

## Workaround we are using

Wrapping `observeElementOffset` so that unsubscribing marks the subscription dead; upstream's orphaned timer still fires but finds nothing to call. It works, but it is a wrapper around a timer we cannot reach, which is why this is filed rather than fixed locally.

Happy to open a PR if the shape above is the one you would want.

solomonaustin:
Fixes #1255. The analysis in that issue is @Zinn-Digital-LTD's and it was accurate down to the line — they offered to send a patch, so if they'd rather own this, I'm glad to close in favour of theirs.

### Problem

`observeOffset` arms a debounce on every `scroll` event to reset `isScrolling` back to `false` after `isScrollingResetDelay` (150 ms default). Its disposer removes the scroll/scrollend listeners but leaves that timer armed, and the handle is local to the `debounce` closure:

```ts
export const debounce = (targetWindow, fn, ms) => {
  let timeoutId: number          // <- unreachable from outside
  return function (this: any, ...args) {
    targetWindow.clearTimeout(timeoutId)
    timeoutId = targetWindow.setTimeout(() => fn.apply(this, args), ms)
  }
}
```

so a consumer cannot clear it either — the fix has to live in the library.

Unmount a virtualizer inside that window and one late `(offset, false)` callback still arrives, running `maybeNotify()` → the consumer's `onChange` → in React a `useReducer` dispatch into a tree that no longer exists. A browser swallows that; under vitest/jsdom it surfaces as an update-after-unmount and can fail a suite that is otherwise correct.

It's the default path rather than an edge case: `useScrollendEvent` defaults to `false`, and jsdom has no `onscrollend`, so every jsdom-based suite goes through the debounce.

### Fix

`debounce` returns a function carrying `cancel()`, and the disposer already returned by `observeOffset` calls it alongside the listener removal. That disposer is pushed onto `unsubs`, which `cleanup()` drains, so element and window observers are both covered and no call site changes.

`debounce` is publicly exported, so this is additive — no existing signature changes.

### Testing

Added `observeElementOffset: cleanup drops the queued isScrolling reset`, which scrolls, tears down inside the window, advances fake timers past the delay and asserts no further callback.

Verified it fails for the right reason without the source change — the orphaned timer fires and the callback lands once (`Number of calls: 1`) — and passes with it.

- `vitest run` in `virtual-core` — 129/129 pass
- `tsc`, `eslint ./src`, `vite build` all clean
- Patch changeset included

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented delayed scroll-state callbacks from firing after virtualized elements are unmounted or observation is stopped.
  * Reset scrolling status and direction during cleanup, preventing stale scroll state when virtualization is disabled, reused, or its scroll element changes.
  * Improved cleanup during scroll observation teardown, reducing unexpected updates after removal.

* **Tests**
  * Added regression coverage for cleanup, scroll-element changes, disabling, unmounting, and cancellation of pending callbacks.

### DIFF
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Cancel the pending `isScrolling` reset when a scroll observer is torn down, so a virtualizer unmounted within `isScrollingResetDelay` of a scroll no longer emits one late `onChange`.
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -2,4 +2,4 @@
 '@tanstack/virtual-core': patch
 ---
 
-Cancel the pending `isScrolling` reset when a scroll observer is torn down, so a virtualizer unmounted within `isScrollingResetDelay` of a scroll no longer emits one late `onChange`.
+Cancel the pending `isScrolling` reset when a scroll observer is torn down, and reset `isScrolling` and `scrollDirection` in `cleanup()` so they don't stay stuck after the scroll element changes or is removed.
--- a/packages/marko-virtual/e2e/app/e2e/option-gates.spec.ts
+++ b/packages/marko-virtual/e2e/app/e2e/option-gates.spec.ts
@@ -101,17 +101,6 @@ test('enabled=false disables the virtualizer (empty window); enabled=true re-win
   const deep = await renderedIndexes(page)
   expect(deep[0]!).toBeGreaterThan(50)
 
-  // Let the end-of-scroll debounce fire while STILL ENABLED before toggling.
-  // KNOWN UPSTREAM CORE BUG (found by this gate): core's debounce (utils.ts) has no
-  // cancel, and observeOffset's unsubscribe only removes the event listeners — a
-  // pending end-of-scroll timer survives cleanup() and later fires
-  // cb(staleOffset, false) into the live instance. Disable + re-enable within
-  // isScrollingResetDelay (150ms) and the stale offset overwrites the correct
-  // re-enable recompute, leaving a stale window until the next real scroll event.
-  // This wait sidesteps the zombie timer so the gate asserts the enabled contract
-  // itself; remove it if/when the core fix (cancellable debounce) lands.
-  await page.waitForTimeout(250)
-
   // Disable: the deep window disappears (measurements cleared, empty/collapsed window).
   await page.locator('[data-testid="toggle"]').click()
   await page.waitForFunction(
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -217,6 +217,10 @@ const observeOffset = <T extends Element | Window>(
     if (registerScrollendEvent) {
       element.removeEventListener('scrollend', endHandler)
     }
+    // Removing the listener doesn't retract a reset already queued by the
+    // last scroll, and that call would land on a virtualizer that has been
+    // torn down — in React, a dispatch into an unmounted tree.
+    fallback?.cancel()
   }
 }
 
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -767,6 +767,13 @@ export class Virtualizer<
       this.rafId = null
     }
     this.scrollState = null
+    // The debounce cancelled above is the only thing that writes `isScrolling`
+    // back to false, so a cleanup inside the reset window would strand it, and
+    // the direction derived from it, as true. That matters because `cleanup`
+    // also runs when the scroll element changes or `enabled` goes false, where
+    // the instance lives on.
+    this.isScrolling = false
+    this.scrollDirection = null
     // The iOS gesture/deferral state is scoped to the current scroll
     // element: the touch listeners that maintain it were just removed, and
     // an in-flight touch keeps targeting the old element (implicit touch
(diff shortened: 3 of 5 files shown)


piecyk (changes requested):
Diagnosis is accurate and the fix is the right shape — `cancel()` in the shared `observeOffset` disposer covers both observers, needs no call-site changes, and is additive on a public util so `patch` is correct. Verified locally: virtual-core 129/129, react-virtual 7/7, `tsc` clean, and the new test does fail without the source change.

One thing to fix first: that debounce is the **only** writer of `isScrolling = false` — it's assigned in just one place, the scroll callback in `_willUpdate()`.

That's harmless on unmount, but `cleanup()` also runs when the scroll element changes and when `enabled` goes `false`, and there the instance stays alive. Scroll → disable → wait past `isScrollingResetDelay` → re-enable now leaves `isScrolling` stuck `true`, where before the PR the zombie timer reset it to `false`. A stuck flag skips sync measurement in `measureElement`, strands iOS deferred adjustments in `_flushIosDeferredIfReady`, and leaves `scrollDirection` stale.

Could you add this to `cleanup()`, next to the existing `scrollState = null`, and cover the swap path in a test as well as unmount?

```ts
this.isScrolling = false
this.scrollDirection = null
```

Also: `packages/marko-virtual/e2e/app/e2e/option-gates.spec.ts` has a `waitForTimeout(250)` with a note to remove it once a cancellable debounce lands. Worth dropping here — that's exactly the disable-within-150ms case above, so it doubles as verification.

Minor: the check rollup only shows CodeRabbit and Socket, so the test workflow doesn't look like it has run yet.

solomonaustin:
Thanks — that's a real gap, and you're right about the mechanism. I went and checked rather than take it on trust: `isScrolling` only has two writers, the field initialiser and the scroll callback in `_willUpdate()`, so cancelling the debounce did remove the only runtime path back to `false`. And since `cleanup()` is reached from `_willUpdate` on both the element swap and the `enabled: false` path, the instance survives and the flag sticks. `_flushIosDeferredIfReady` bails early on `isScrolling` and the measure paths gate on it too, so the knock-on effects are as you described.

Added both resets in `cleanup()` next to `scrollState = null`, with a short note. The iOS block right underneath already documents the same hazard — an unsub clearing the only pending reset of a flag — so it seemed right to keep them together.

Three tests, one per way into `cleanup()`: element swapped, element removed, unmount. Each drives the instance mid-scroll through the offset callback and then asserts both flags. All three fail without the source change.

I also dropped the `waitForTimeout(250)` in the marko option-gates gate. One correction there though: I checked whether it doubles as verification like you suggested, and it doesn't. I built the packages and ran the spec — 11/11 — then reverted the `cleanup()` change, rebuilt, and ran it again. Still 11/11, gate 2 included. So it isn't sensitive to this bug either way. Worth removing as a stale workaround, and it saves 250ms, but the unit tests are the actual guard here.

Changeset now mentions the flag reset as well, since that's user-visible too.

virtual-core 132/132, react-virtual 7/7, marko e2e 11/11, `tsc` and `eslint` clean.

On the checks — you're right they haven't run. I can't trigger the workflow from a fork, so I think it needs an approval from your side.

piecyk (approved):

Issue #1258, opened 2026-08-24
paddingEnd option breaks followOnAppend + anchorTo: 'end'


GianlucaGuarini:
### Describe the bug

When `paddingEnd` is set alongside `anchorTo: 'end'` and `followOnAppend`, the follow-on-append behaviour silently stops working. The virtualizer never scrolls to the end after new items are appended, even when the viewport was already pinned at the bottom.

The root cause is an internal inconsistency between two "distance from end" calculations:

**`getDistanceFromEnd()`** — used by `isAtEnd()` and the `followOnAppend` trigger in `setOptions`:

```ts
// reads real DOM scrollHeight - clientHeight
return Math.max(this.getMaxScrollOffset() - this.getScrollOffset(), 0)
```

**`getVirtualDistanceFromEnd()`** — used by `resizeItem`'s end-anchor guard (`wasAtEnd`):

```ts
// uses getTotalSize(), which adds paddingEnd
return Math.max(this.getTotalSize() - this.getSize() - this.getScrollOffset(), 0)
```

`getTotalSize()` includes `paddingEnd`:

```ts
return Math.max(end - this.options.scrollMargin + this.options.paddingEnd, 0)
```

So `getVirtualDistanceFromEnd()` always reports a distance of **at least `paddingEnd` pixels**, even when the user is scrolled to the absolute bottom of the container.

The `followOnAppend` trigger in `setOptions` calls `this.isAtEnd(prevOptions.scrollEndThreshold)` — this uses the DOM-based path and correctly detects the pinned state. So `followOnAppend` *is* triggered and `scrollToEnd()` *is* scheduled.

But the `wasAtEnd` guard in `resizeItem` uses `getVirtualDistanceFromEnd()`:

```ts
const wasAtEnd =
  this.options.anchorTo === 'end' &&
  this.scrollState?.behavior !== 'smooth' &&
  this.getVirtualDistanceFromEnd() <= this.options.scrollEndThreshold
```

With e.g. `paddingEnd: 80` and `scrollEndThreshold: 80`, `getVirtualDistanceFromEnd()` returns `>= 80` even at the very bottom, so `wasAtEnd` is always `false`. The end-anchor scroll adjustment in `resizeItem` never fires while a streaming message is growing, the viewport drifts away from the bottom, and by the time `scrollToEnd()` runs it either undershoots or the user is already well above the end.

### Your minimal, reproducible example

https://github.com/tanstack/virtual/tree/main/examples/react/chat

[...]

PR #1265, opened 2026-09-07
fix(react-virtual): size direct DOM container before scroll adjustments

### LINKED ISSUE #1266: [React] directDomUpdates loses end anchoring when an earlier row grows
### Describe the bug

With `anchorTo: 'end'` and `directDomUpdates: true`, growing an earlier row can leave the viewport above the bottom even though it was previously pinned. The gap persists when the last row keeps its size.

The standalone reproduction uses eight messages with exact initial size estimates. The last row is 300px tall, matching the viewport, and only the preceding row grows by 24px. The browser reports a 24px bottom gap after the resize.

[Run the live reproduction](https://tigerbea.github.io/tanstack-virtual-resize-repro/)

[Source and local run instructions](https://github.com/tigerBeA/tanstack-virtual-resize-repro)

The example uses published npm packages: `@tanstack/react-virtual@3.14.10`, `@tanstack/virtual-core@3.17.8`, and React 19.2.8. It sets `useFlushSync: false` and uses native ResizeObserver measurements. CSS scroll anchoring is disabled with `overflow-anchor: none` to isolate the virtualizer's adjustment.

### Steps to reproduce

1. Open the preview and wait for **Actual bottom gap: 0 px**.
2. Click **Grow previous message by 24 px** once.
3. Observe **Actual bottom gap: 24 px**. The last row's bottom is now below the viewport.
4. Reload to repeat.

Measured from the actual scroll element:

| DOM value | Before | After |
| --- | ---: | ---: |
| `scrollHeight` | 650 | 674 |
| `clientHeight` | 300 | 300 |
| `scrollTop` | 350 | 350 |
| `scrollHeight - clientHeight - scrollTop` | 0 | 24 |

The diagnostics are rendered by a separate sibling component, so updating them does not rerender the virtualized list.

### Expected behavior

The viewport remains pinned: `scrollTop` advances to 374 and the bottom gap stays at 0px when the preceding row grows.

Every time with the steps above.

### Screenshots or Videos

After growing the previous row: actual bottom gap is 24px

### Platform

- macOS 26.5.2
- Chromium 141.0.7390.37

### tanstack-virtual version

`@tanstack/react-virtual@3.14.10` / `@tanstack/virtual-core@3.17.8`

### TypeScript version

Not applicable; the reproduction uses JavaScript/JSX.

### Additional context

`resizeItem` updates the measurements and applies its scroll adjustment before `onChange` updates the React adapter's direct DOM size container. The browser clamps the adjustment against the old `scrollHeight`; the container grows afterwards, leaving the actual DOM offset behind. Keeping the last row unchanged makes the gap persist.

Proposed fix: #1265. It applies the adapter's existing `applyContainerSize` before delegating to `scrollToFn`, with a browser regression test. This follows the ordering principle used for prepend synchronization in #1237, and covers the resize path. This reproduction does not use `paddingEnd`, unlike #1258.

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct.
- [x] I understand that if my bug cannot be reliably reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

tigerBeA:
## 🎯 Changes

Fixes #1266.

[Live reproduction](https://tigerbea.github.io/tanstack-virtual-resize-repro/) · [Standalone source](https://github.com/tigerBeA/tanstack-virtual-resize-repro)

With `anchorTo: 'end'` and `directDomUpdates`, growing an earlier message can leave the viewport above the bottom when the last message keeps its size. `resizeItem` applies its scroll adjustment before notifying the React adapter, so the browser clamps the write against the old sizer height. The internal offset advances, but the actual DOM position remains behind. In the regression, growing the previous message by 24px leaves a persistent 24px gap.

Apply the adapter's existing, size-guarded `applyContainerSize` before delegating to `scrollToFn`. The browser then sees the updated scroll range when applying the adjustment. This covers the resize path in direct DOM mode using the same ordering principle as #1237; related to #1209. The change is scoped to the React adapter's direct DOM mode.

A separate, statically configured chat-resize E2E fixture uses `directDomUpdates` with `useFlushSync: false`; the original chat fixture and its default-mode coverage are preserved. The regression grows the previous message while keeping the last message at a fixed viewport height, then checks the real `scrollHeight` and `scrollTop`. It uses the native ResizeObserver and browser scrolling, with native scroll anchoring disabled so it cannot mask a missing virtualizer adjustment.

Validation (Node 24.8.0 from `.nvmrc`, pnpm 11.9.0 from `packageManager`):

- With only the SDK fix removed, the final `/chat-resize/` regression fails with an actual bottom gap of 24px. Restoring the fix makes it pass.
- `NX_NO_CLOUD=true NX_DAEMON=false CI=1 pnpm run test:pr --parallel=2 --skip-nx-cache` passes, including all 34 React browser tests, 7 React library tests, type checks, lint, builds, and package validation. Nx caching was disabled for this run.
- Prettier checks pass for all changed files.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a patch changeset for `@tanstack/react-virtual`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed end-anchored virtualized lists so they remain pinned to the bottom when an earlier item increases in height.
  - Updated scroll adjustments to use the latest container size, preventing the browser from limiting the intended scroll position.
  - Improved chat-style scrolling so users remain at the latest message after earlier content expands.

- **Tests**
  - Added end-to-end coverage for chat-style lists where a previous message grows while scrolled to the bottom.

### DIFF
--- a/.changeset/fresh-chats-scroll.md
+++ b/.changeset/fresh-chats-scroll.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-virtual': patch
+---
+
+Update the direct DOM size container before applying scroll adjustments, so resizing an earlier item keeps an end-anchored list pinned even when the last item does not resize.
--- a/packages/react-virtual/e2e/app/chat-resize/index.html
+++ b/packages/react-virtual/e2e/app/chat-resize/index.html
@@ -1,0 +1,10 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="./main.tsx"></script>
+  </body>
+</html>
--- a/packages/react-virtual/e2e/app/chat-resize/main.tsx
+++ b/packages/react-virtual/e2e/app/chat-resize/main.tsx
@@ -1,0 +1,84 @@
+import React from 'react'
+import { createRoot } from 'react-dom/client'
+import { useVirtualizer } from '@tanstack/react-virtual'
+
+const VIEWPORT_HEIGHT = 300
+const initialMessages = Array.from({ length: 8 }, (_, index) => ({
+  id: `m-${index}`,
+  height: index === 7 ? VIEWPORT_HEIGHT : 50,
+}))
+
+function App() {
+  const [messages, setMessages] = React.useState(initialMessages)
+  const parentRef = React.useRef<HTMLDivElement>(null)
+  const virtualizer = useVirtualizer({
+    count: messages.length,
+    getScrollElement: () => parentRef.current,
+    getItemKey: (index) => messages[index]!.id,
+    estimateSize: (index) => initialMessages[index]!.height,
+    anchorTo: 'end',
+    followOnAppend: true,
+    scrollEndThreshold: 4,
+    overscan: 4,
+    directDomUpdates: true,
+    useFlushSync: false,
+  })
+
+  React.useLayoutEffect(() => {
+    virtualizer.scrollToEnd()
+  }, [virtualizer])
+
+  return (
+    <div>
+      <button
+        id="grow-previous"
+        onClick={() => {
+          setMessages((current) =>
+            current.map((message, index) =>
+              index === current.length - 2
+                ? { ...message, height: message.height + 24 }
+                : message,
+            ),
+          )
+        }}
+      >
+        Grow previous
+      </button>
+      <div
+        ref={parentRef}
+        id="scroll-container"
+        style={{
+          height: VIEWPORT_HEIGHT,
+          width: 420,
+          overflow: 'auto',
+          overflowAnchor: 'none',
+        }}
+      >
+        <div
+          ref={virtualizer.containerRef}
+          style={{ position: 'relative', width: '100%' }}
+        >
+          {virtualizer.getVirtualItems().map((item) => (
+            <div
+              key={item.key}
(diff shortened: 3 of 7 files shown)


piecyk:
@tigerBeA Thanks for this, and especially for the reproduction. It made the root cause unambiguous, and your `chat-resize` app and test are now in #1260 as the permanent regression test for this scenario, credited to you in both files.

I'm closing this in favour of #1260, which fixes #1258 and #1266 together in virtual-core. The two issues are the same mechanism: `resizeItem` writes the compensating `scrollTop` before the consumer has committed the new total size, the browser clamps the write to the old maximum, and nothing re-applies the lost distance afterwards. Your PR prevents the clamp for the React `directDomUpdates` path by sizing the container inside `scrollToFn`, which works, but only for that one adapter and mode. The core fix records a compensation write whose target exceeds the scroll maximum at write time and re-issues it once the sizer has grown, so it also covers the React-rendered sizer, `useWindowVirtualizer`, and the other adapters, and it recovers even when no re-render follows the resize.

Your diagnosis in #1266 was spot on, including the ordering point about `applyContainerSize`. Sizing the container before the write is still a reasonable optimisation for the direct-DOM path, since it avoids the extra write entirely, so if you'd like to rebase that part on top of #1260 once it lands I'd be happy to review it as a follow-up.

PR #1272, opened 2026-09-08
fix(virtual-core): preserve measurement keys and follow sliding windows

### LINKED ISSUE #1270: [virtual-core] Lazy measurement keys become inconsistent after a same-count data update
### Describe the bug

In the single-lane measurement path, a stable `getItemKey` callback reading updated data can leave rendered item keys inconsistent with the current message IDs after a same-count update. With `anchorTo: 'end'`, the same update can also lose the reading anchor even with `followOnAppend: false`.

The reproduction starts with messages 1–20 and Message 9 at the viewport top. Replacing the array with messages 2–21 leaves `scrollTop` at 400 instead of 350. The rendered range contains 8 key/ID mismatches; for example, Message 10 has `VirtualItem.key` 9. Message 9 moves 50 px above the viewport top.

[Live cache reproduction](https://tigerbea.github.io/tanstack-virtual-sliding-window-repro/cache.html)

[Pinned source](https://github.com/tigerBeA/tanstack-virtual-sliding-window-repro/blob/7c5088f7d20bbeed9ab04f50a03ed5a0d58b0f0f/src/cache.js) · [Source and local run instructions](https://github.com/tigerBeA/tanstack-virtual-sliding-window-repro/tree/7c5088f7d20bbeed9ab04f50a03ed5a0d58b0f0f)

The page uses published `@tanstack/virtual-core@3.17.9` directly, with the standard element observers and scrolling function. It renders only the virtual range, including overscan, and uses exact 50 px row estimates in a 300 px viewport. Native CSS scroll anchoring is disabled. No network, dynamic row measurements, or custom scroll compensation is involved.

### Steps to reproduce

1. Open the cache page. Confirm **Cached keys different from data IDs: 0**, `scrollTop: 400`, and Message 9 at the viewport top.
2. Click **Trim oldest + append** once. The array is replaced with messages 2–21, then `setOptions` receives the new options. The count and `getItemKey` function identity stay the same; the callback reads the current array.
3. Observe **8** key/ID mismatches and `scrollTop: 400`. Message 10 is now at the viewport top; its cached key is 9.
4. Click **Reset cache example** to repeat.

### Expected behavior

After the update, `VirtualItem.key` should match the item represented by its index. The reading anchor should remain Message 9 at the viewport top, which requires `scrollTop: 350` after the oldest 50 px row is removed. This expectation does not require following appended output; `followOnAppend` is explicitly disabled.

Every time with the steps above.

### Platform

- macOS 26.5.2
- Chrome 152 (reported user-agent major version)
- Verified on 2026-09-08 using a production Vite build

### tanstack-virtual version

`@tanstack/virtual-core@3.17.9`

### TypeScript version

Not applicable; the reproduction is JavaScript.

### Additional context

Proposed fix: #1272.

Related: #1271 describes append-and-trim updates losing end following. This cache reproduction disables following and demonstrates a separate key-consistency failure.

The single-lane layout records positions and sizes immediately, but unread lazy items resolve their keys later through `getItemKey`. If the data changes between those steps, materialized items retain their old keys while unread items obtain keys from the new array. In this fixture, `setOptions` reads the old first/last keys lazily after the array has changed, so its edge comparison misses the data update.

The suspected invariant is that a measurement's key should belong to the same layout build as its position and size. The fixture does not read the private measurement cache or pre-read the first/last items to influence change detection.

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct.
- [x] I understand that if my bug cannot be reliably reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

### LINKED ISSUE #1271: [virtual-core] followOnAppend loses end position when appending and trimming together
### Describe the bug

With `anchorTo: 'end'` and `followOnAppend: true`, appending a new item while trimming the oldest item in the same update moves an end-pinned viewport away from the bottom when the count stays unchanged.

The reproduction starts with 20 messages, each 50 px tall, in a 300 px viewport. Changing messages 1–20 to 2–21 keeps the total height at 1000 px, but moves `scrollTop` from 700 to 650, leaving a 50 px bottom gap. Appending without trimming follows correctly.

[Live window-follow reproduction](https://tigerbea.github.io/tanstack-virtual-sliding-window-repro/)

[Pinned source](https://github.com/tigerBeA/tanstack-virtual-sliding-window-repro/blob/7c5088f7d20bbeed9ab04f50a03ed5a0d58b0f0f/src/main.jsx) · [Source and local run instructions](https://github.com/tigerBeA/tanstack-virtual-sliding-window-repro/tree/7c5088f7d20bbeed9ab04f50a03ed5a0d58b0f0f)

Published packages: `@tanstack/react-virtual@3.14.11`, `@tanstack/virtual-core@3.17.9`, and React 19.2.8. Message IDs are persistent, and the memoized `getItemKey` callback captures each immutable message array. Native CSS scroll anchoring is disabled. There is no streaming, network, dynamic measurement, or custom scroll compensation.

### Steps to reproduce

1. Open the page and confirm **Actual bottom gap: 0.00 px**.
2. Click **Append + trim oldest** once. The array changes from messages 1–20 to 2–21 in one state update.
3. Observe **Actual bottom gap: 50.00 px**. The latest message is below the viewport.
4. Click **Reset** to repeat.
5. As a control, reset and click **Append only (control)**. The count grows to 21 and the bottom gap remains 0 px.

### Expected behavior

A viewport already within `scrollEndThreshold` of the end should stay at the end when new items are appended and older items are trimmed in the same update. In this example `scrollTop` should remain 700 and the bottom gap should remain 0.

A user reading earlier content should retain their reading position. For these non-growing updates, a replacement, reorder, or trim without new appended items should not be treated as appended output merely because the last key changes. This report does not request a change to the existing count-growing classification.

Every time with the steps above.

### Screenshots or Videos

Append and trim leaves a 50 px bottom gap

Measured from the real scroll element:

| Scenario | Count | scrollHeight | clientHeight | scrollTop | Bottom gap |
| --- | ---: | ---: | ---: | ---: | ---: |
| Initial | 20 | 1000 | 300 | 700 | 0 |
| Append + trim | 20 | 1000 | 300 | 650 | 50 |
| Append only, after reset | 21 | 1050 | 300 | 750 | 0 |

### Platform

- macOS 26.5.2
- Chrome 152 (reported user-agent major version)
- Verified on 2026-09-08 using a production Vite build

### tanstack-virtual version

`@tanstack/react-virtual@3.14.11` / `@tanstack/virtual-core@3.17.9`

### TypeScript version

Not applicable; the reproduction uses JavaScript/JSX.

### Additional context

Proposed fix: #1272.

Related: #1270 describes inconsistent lazy measurement keys with a stable callback. This reproduction uses a callback that captures each immutable array and still loses end following, so the two behaviors can be reproduced independently.

The core's append-follow condition in `setOptions` requires `nextCount > prevCount`. This update changes the edge keys without increasing the count, so it takes the visible-item anchoring path instead of following the new tail.

The reproduction retains an ordered, non-empty suffix of the old list and appends a new key. It does not rely on treating wholesale replacement or recycled keys as an append.

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct.
- [x] I understand that if my bug cannot be reliably reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

tigerBeA:
## 🎯 Changes

Fixes #1270.
Fixes #1271.

Keep measurement keys consistent across data updates, then use those keys to follow appended output when older items are trimmed in the same update.

Both failures have standalone live reproductions using published packages:

| Reproduction | Published core 3.17.9 | Local candidate |
| --- | --- | --- |
| [Cached item identity](https://tigerbea.github.io/tanstack-virtual-sliding-window-repro/cache.html), following disabled | 8 rendered key/ID mismatches; reading anchor moves by -50 px | 0 mismatches; reading anchor stays in place |
| [Append + trim](https://tigerbea.github.io/tanstack-virtual-sliding-window-repro/), initially at the end | 50 px bottom gap | 0 px bottom gap |

[Reproduction source and steps](https://github.com/tigerBeA/tanstack-virtual-sliding-window-repro/tree/7c5088f7d20bbeed9ab04f50a03ed5a0d58b0f0f).

The implementation is split into two commits for review:

1. **Preserve keys in lazy measurement builds.** Single-lane cache slots initially hold the keys used to construct the layout, then hold `VirtualItem` objects after first access. Partial rebuilds retain the untouched prefix. Unread items no longer call a stable `getItemKey` against newer data, so old edge comparisons and reading anchors use the correct identities.
2. **Follow append-and-trim updates.** For non-growing updates, recognize an ordered, non-empty suffix of the old list followed by new keys. These non-growing updates are not followed when they are replacements, reorders, rotations, trim-only updates, or prepends. The existing count-growing classification is unchanged; this PR does not add key-continuity validation to that path. Following still requires the viewport to have been within `scrollEndThreshold`. When following wins, avoid eagerly assigning the reading-anchor offset: an equal-height window update can leave the DOM offset unchanged and emit no scroll event to correct that assignment.

No public options are added. One patch changeset covers `@tanstack/virtual-core`, and the `followOnAppend` documentation describes the retained-overlap requirement.

### Cost

Keys use the existing lazy-cache slots rather than a parallel key array; `VirtualItem` objects remain lazy and the flat position/size buffer is reused. The overlap scan reads raw keys without materializing all old rows and runs only for end-pinned, non-growing append candidates. The tradeoff is storing keys during the layout loop and copying cache slots for partial rebuilds.

See the [performance measurements and rerun instructions](https://github.com/tigerBeA/tanstack-virtual-sliding-window-repro/blob/667bf85512880cba915e994c9ec4fa7d825eb947/performance/README.md) for comparison sources and recorded resize, memory, browser-update, and native-scroll results. At 10,000 rows, head-heavy rebuilds added about 12–16 µs per operation (8.5–11.5%); the ordinary-scroll samples showed no consistent slowdown. The baseline is an earlier local window-follow fix, not the unmodified release. Browser timings exclude paint.

### Validation

- With the original core source and the new cache tests, all 3 cache regressions fail: unread keys resolve to the next item's key, and the reading anchor receives no adjustment.
- With only the first commit, 7 window-follow cases fail; restoring the second commit makes them pass.
- 60 targeted core tests pass, covering lazy keys, partial rebuilds, end anchoring, consecutive window updates, thresholds, disabled following, smooth following, count growth/reduction, and non-append controls. The growing-window compatibility case removes 1 item and appends 2; it passes against both the upstream base and this PR. No persistent skips were added.
- Chrome production-build checks confirm the live failures and the local candidate results above. Ordinary append still follows, and native scrolling away from the end retains the reading anchor on the next window update.
- Prettier checks pass for the changed TypeScript files and changeset.
- Before the latest test-only follow-up, `NX_NO_CLOUD=true NX_DAEMON=false NX_BASE=origin/main NX_HEAD=HEAD CI=1 pnpm run test:pr --parallel=2 --skip-nx-cache` passed for all 69 affected projects at `c452d482`, including 161 core tests, browser tests, type checks, lint, documentation/package checks, and builds. The follow-up adds only one test parameter; production source is unchanged, and the 60-test targeted run above covers the updated file. Runtime: Node 24.8.0 from `.nvmrc`, pnpm 11.9.0 from `packageManager`.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a patch changeset for `@tanstack/virtual-core`.

## Summary by CodeRabbit

- **Bug Fixes**
  - End-pinned virtualizers now continue following newly appended items when older items are trimmed in the same update.
  - Previously measured item keys remain stable when underlying data changes, preventing rows from unexpectedly changing identity.
  - Improved reading-anchor preservation across sliding-window updates, including smooth scrolling and updates without scroll events.

- **Documentation**
  - Clarified requirements and limitations for following appended items during sliding-window updates, including persistent keys, retained items, and cases that are not automatically followed.

### DIFF
--- a/.changeset/follow-sliding-window.md
+++ b/.changeset/follow-sliding-window.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/virtual-core': patch
+---
+
+Keep an end-pinned virtualizer following appended items when older items are trimmed in the same update and the item count does not increase. Recognize ordered, overlapping windows while preserving reading anchors for users who have scrolled away from the end.
+
+Preserve item keys in the lazy measurement cache so a stable `getItemKey` callback reading mutable data cannot change the identity of previously measured rows.
--- a/docs/api/virtualizer.md
+++ b/docs/api/virtualizer.md
@@ -271,6 +271,8 @@ When used with `anchorTo: 'end'`, controls whether the virtualizer scrolls to th
 
 Passing `true` is equivalent to `'auto'`. Passing a scroll behavior uses that behavior for the follow.
 
+Following also works when older items are trimmed from the start in the same update without increasing the count. This requires persistent keys, a non-empty suffix of the old list retained in order, and appended items with new keys. Non-growing updates with no retained items are not automatically followed.
+
 This option does not follow prepends. It only follows appended output, and only when the viewport was already within `scrollEndThreshold` of the end before the append.
 
 ### `scrollEndThreshold`
--- a/packages/virtual-core/src/index.ts
+++ b/packages/virtual-core/src/index.ts
@@ -409,9 +409,12 @@ export class Virtualizer<
   isScrolling = false
   private scrollState: ScrollState | null = null
   measurementsCache: Array<VirtualItem> = []
-  // Flat backing store for the lanes===1 fast path: [start_0, size_0, start_1, size_1, ...].
-  // null until the first single-lane build; reused (and grown) across rebuilds.
-  private _flatMeasurements: Float64Array | null = null
+  // Keys belong to the layout build, even when VirtualItems are read later.
+  // The flat [start, size, ...] buffer is reused across builds.
+  private _singleLaneMeasurements: {
+    flat: Float64Array
+    items: Array<Key | VirtualItem>
+  } | null = null
   itemSizeCache = new Map<Key, number>()
   private itemSizeCacheVersion = 0
   private laneAssignments = new Map<number, number>() // index → lane cache
@@ -1272,6 +1275,7 @@ export class Virtualizer<
       const itemSizeCache = this.itemSizeCache
       if (!enabled) {
         this.measurementsCache = []
+        this._singleLaneMeasurements = null
         this.itemSizeCache.clear()
         this.laneAssignments.clear()
         return []
@@ -1291,6 +1295,7 @@ export class Virtualizer<
         this.lanesChangedFlag = false // Reset immediately
         this.lanesSettling = true // Start settling period
         this.measurementsCache = []
+        this._singleLaneMeasurements = null
         this.itemSizeCache.clear()
         this.laneAssignments.clear() // Clear lane cache for new lane count
         // Force min = 0 on the rebuild
@@ -1320,20 +1325,21 @@ export class Virtualizer<
       // per-item VirtualItem object allocation. We write start/size pairs
       // into a Float64Array and return a Proxy that builds VirtualItem
       // objects on demand (only the indices a consumer actually reads).
-      //
-      // At n=100k this drops cold-mount cost from ~2.5ms (eager object
-      // allocation) to roughly the cost of a single typed-array fill.
       if (lanes === 1) {
         // Reuse flat backing if large enough; else grow (preserving data
         // before `min` to mirror the slice-and-rebuild contract).
         const need = count * 2
-        let flat = this._flatMeasurements
+        let flat = this._singleLaneMeasurements?.flat
         if (!flat || flat.length < need) {
           const next = new Float64Array(need)
           if (flat && min > 0) next.set(flat.subarray(0, min * 2))
           flat = next
-          this._flatMeasurements = flat
         }
+
+        const items: Array<Key | VirtualItem> =
+          min === 0
+            ? new Array(count)
+            : this._singleLaneMeasurements!.items.slice()
 
         let runningStart: number
         if (min === 0) {
@@ -1346,6 +1352,7 @@ export class Virtualizer<
 
         for (let i = min; i < count; i++) {
           const key = getItemKey(i)
+          items[i] = key
           const measuredSize = itemSizeCache.get(key)
           const size =
             typeof measuredSize === 'number'
@@ -1356,7 +1363,8 @@ export class Virtualizer<
           runningStart += size + gap
         }
 
-        const view = createLazyMeasurementsView(count, flat, getItemKey)
+        this._singleLaneMeasurements = { flat, items }
+        const view = createLazyMeasurementsView(items, flat)
         this.measurementsCache = view
         return view
       }
@@ -1490,8 +1498,8 @@ export class Virtualizer<
         lanes,
         // Pass the typed array so binary search + forward-walk can read
         // start/end directly from Float64Array, skipping the Proxy traps.
-        lanes === 1 && this._flatMeasurements != null
-          ? this._flatMeasurements
+        lanes === 1 && this._singleLaneMeasurements !== null
+          ? this._singleLaneMeasurements.flat
           : null,
       )
       return this.range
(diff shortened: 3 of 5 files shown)


tigerBeA:
@coderabbitai full review

piecyk (approved):
Thanks, this is a solid pair of fixes and the reproductions made it easy to verify.

PR #1274, opened 2026-09-09
fix(angular-virtual): remove deprecated effect option

### LINKED ISSUE #1058: Angular - console warn "allowSignalWrites"
### Describe the bug

Angular - 20.2.4, console warn "The 'allowSignalWrites' flag is deprecated and no longer impacts effect() (writes are always allowed)".

https://stackblitz.com/edit/tanstack-virtual-bmut3piq

### Steps to reproduce

1. open stackblitz
2. open developer tools, console tab

### Expected behavior

no warn show

Every time

### Screenshots or Videos

Image

### Platform

Windows 11

### tanstack-virtual version

v4.0.0

### TypeScript version

~5.9.2

### Additional context

_No response_

### Terms & Code of Conduct

ousamabenyounes:
## Changes

Fixes #1058.

### Problem

Angular 20 logs a deprecation warning when the affected examples load because they still pass `allowSignalWrites: true` to `effect`.

### Fix

- Removed the deprecated effect option from the Angular padding, dynamic, and infinite-scroll examples.
- Added an Angular e2e guard that fails if source examples reintroduce `allowSignalWrites: true`.

## Test verification (RED -> GREEN)

RED:

- Browser verification against the Angular infinite-scroll example logged the deprecated `allowSignalWrites` warning before the fix.
- The new guard was run with the example edits reverted and failed, listing all eight affected Angular source files.

GREEN:

- Browser verification against the same example no longer logged the deprecated warning after the fix.
- The new guard passes with the fix applied.

## Full local suite proof

- Baseline upstream `main`: `pnpm run test:ci` passed.
- Final patched tree: `pnpm exec nx run-many --parallel=1 --targets=test:sherif,test:knip,test:docs,test:eslint,test:lib,test:e2e,test:types,test:build,build` passed.

The final full local suite used Nx parallelism of 1 to avoid local browser-resource flakes seen during the high-concurrency attempt. The command covers the same target matrix as `pnpm run test:ci`.

## Release impact

- [ ] Published package code changed
- [x] Examples or tests only

No changeset is included because this only updates examples and test coverage.

## Summary by CodeRabbit

- **Refactor**
  - Updated Angular examples to use default effect options while preserving existing measurement and data-fetching behavior.
  - Simplified effect declarations for improved readability.

- **Tests**
  - Added end-to-end coverage that detects deprecated effect options in Angular example files.

### DIFF
--- a/examples/angular/dynamic/src/app/column-virtualizer-dynamic.component.ts
+++ b/examples/angular/dynamic/src/app/column-virtualizer-dynamic.component.ts
@@ -56,12 +56,10 @@ export class ColumnVirtualizerDynamic {
 
   count = this.sentences.length
 
-  #measureItems = effect(
-    () =>
-      this.virtualItems().forEach((el) => {
-        this.virtualizer.measureElement(el.nativeElement)
-      }),
-    { allowSignalWrites: true },
+  #measureItems = effect(() =>
+    this.virtualItems().forEach((el) => {
+      this.virtualizer.measureElement(el.nativeElement)
+    }),
   )
 
   virtualizer = injectVirtualizer(() => ({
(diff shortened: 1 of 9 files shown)


piecyk (approved):
Thanks, confirmed #1058 on Angular 20 and that no `allowSignalWrites` usages remain. The guard test is heavier than it needs to be for a source check, but it's harmless, so merging as-is.

PR #1263, opened 2026-09-04
fix(virtual-core): scrollToIndex(last) overshoots when paddingEnd > 0

dikshit-n:
## Summary
Fixes `scrollToIndex(last, { align: 'end' })` so it scrolls to the virtual max offset (content end) rather than the raw DOM max scroll offset (content + padding end) when `paddingEnd > 0`.

## Problem
When `paddingEnd` is set on a virtualizer, `getOffsetForIndex(last, 'end')` returns `getMaxScrollOffset()` = `scrollHeight - clientHeight`. Since `scrollHeight` includes `paddingEnd`, the returned offset overshoots the rendered end of the last item by exactly `paddingEnd` pixels. This makes `scrollToIndex(last)` hide the last item below the viewport when `paddingEnd > 0`.

Closes #1257

## Solution
For the last item with `align: 'end'`, use `Math.max(getTotalSize() - paddingEnd - getSize(), 0)` instead of `getMaxScrollOffset()`. This derives the correct virtual max offset from the virtualizer's own size model (`getTotalSize()` includes `paddingEnd` but we must subtract it back out to get the content-only max scroll). The result keeps the last item's rendered end flush with the bottom of the viewport.

## Changes Made
- `packages/virtual-core/src/index.ts`: In `getOffsetForIndex()`, replaced the raw DOM-based `getMaxScrollOffset()` return for the last item with the virtual-size-based formula that excludes `paddingEnd`.
- `packages/virtual-core/tests/index.test.ts`: Added regression test `#1257: scrollToIndex(last) with paddingEnd keeps the last item flush with the viewport bottom`.

## Testing
- All 133 tests in `@tanstack/virtual-core` pass (1 new + 132 existing).
- Type checks pass (`pnpm nx test:types @tanstack/virtual-core`).
- Build passes (`pnpm nx test:build @tanstack/virtual-core`).
- ESLint passes (`pnpm nx test:eslint @tanstack/virtual-core`).

## Checklist
- [x] Tests pass locally (`pnpm nx test:lib @tanstack/virtual-core`)
- [x] Lint passes
- [x] Code follows repo style
- [x] Documentation updated if needed (N/A — no public API change)
- [x] No console.log or debug code left
- [x] No unrelated changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected end-aligned scrolling to the last item in multi-lane layouts, including shorter lanes.
  * Applied trailing scroll padding consistently when scrolling to the last item.
[...]

Issue #1001, opened 2025-05-13
3.13.8 doesn't scroll all the way to the bottom correctly


iscekic:
### Describe the bug

Exact same code in both cases:
```typescript
// somewhere above
const measureElement: Parameters<typeof useVirtualizer>[0]["measureElement"] = (
  element,
) => element.getBoundingClientRect().height;

const estimateSize = () => 72;

const virtualizer = useVirtualizer({
    count: messages?.length || 0,
    getScrollElement: () => parentRef.current,
    estimateSize,
    measureElement,
    getItemKey: (index: number) =>
      messages?.[index] ? idExtractor(messages[index]) : nanoid(),
  });

// somewhere below
const lastBeforeDateIndex = arr.length - 1; // this is a bit more complicated in the actual code, but most often resolves to this

// some other irrelevant logic, including a getVirtualItemForOffset call

virtualizer.scrollToIndex(Math.max(lastBeforeDateIndex, 0), {
  align: "start",
  behavior: "smooth",
});
```

For context, I have an infinite list in which items (messages) are occasionally appended. The items themselves have a dynamic size. When an item is appended, the list should be scrolled all the way to the bottom, so that the last item is visible.

### Your minimal, reproducible example

/

### Steps to reproduce

/

### Expected behavior

`3.13.8` retains the same behavior as `3.13.7`

### How often does this bug happen?

Every time

### Screenshots or Videos

`3.13.7` - scrolls correctly, bottom of last item is visible

![Image](https://github.com/user-attachments/assets/69414bd0-3eb8-41f4-9c81-cf453495d2cb)

`3.13.8` - doesn't scroll correctly, last item is cut off

![Image](https://github.com/user-attachments/assets/788b1ef0-d6b0-4bb1-b082-6cc62c86a15e)

### Platform

Brave on Linux, though it seems to reproduce everywhere

### tanstack-virtual version

v3.13.8

### TypeScript version

v5.8.3

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

piecyk:
Thanks @iscekic for reporting the issue! I haven’t checked it yet, but it seems related to the loosened approxEqual logic introduced in https://github.com/TanStack/virtual/commit/60719f61b589d6f9d886e4f7c093217f6d693faf

beliarh:
[...]

PR #1276, opened 2026-09-10
fix(virtual-core): use getMaxScrollOffset() - paddingEnd for scrollToIndex(last) end-align

### LINKED ISSUE #1257: `useVirtualizer({paddingEnd: 800})` creates overscroll issue with `scrollToIndex(last)`
### Describe the bug

The `paddingEnd` option for the `useVirtualizer` callback is not correctly taken into account when using `rowVirtualizer .scrollToIndex(last)`. Instead the function call scrolls to the end of the scroll container. For every other index this works as intended. 

```tsx
  count: 100,
  estimateSize: () => 35,
  overscan: 5,
  paddingEnd: 800
})

// ...

<button onClick={() => {
  rowVirtualizer.scrollToIndex(98)
  }}>
  Scroll to 98 (works)
</button>
<button onClick={() => {
  rowVirtualizer.scrollToIndex(99)
  }}>
  Scroll to 99 (overscrolls)
</button>
```

https://github.com/Freymaurer/tanstack-virtual-padding-overscroll-bug-example

### Steps to reproduce

1. Clone and `npm i` -> `npm run dev` the example (fresh vite-react-ts with tanstack virtual dependency) -> open the link printed to console
2. Click the buttons i added on top of the view
  - "Back to Start" uses `scrollToIndex(0)`
  - "Scroll to 98 (works)" scrolls to 98, so not last index
  - "Scroll to 99 (overscrolls)" scrolls to the last index
3. Click on "Back to Start", then "Scroll to 98 (works)". It scrolls with pretty good alignment. The field is shown on the bottom of the scrollContainer
4. Click on "Back to Start", then "Scroll to 99 (overscrolls)". It overscrolls, the element is not visible on screen. 🐛 I think this is due the library calling more or less "scrollToEnd" instead of "scrollTo" in this scenario.

### Expected behavior

I expected the behavior to be consistent. The element at last index should be fully visible after calling the scroll function, instead of scrolling to the scroll container end

Every time

## After clicking on "Scroll to 98 (works)"

Image

## After clicking on "Scroll to 99 (overscrolls)"

Image

### Platform

OS: Windows 11
Browser: Tested on Chrome, Edge, Firefox, Opera

### tanstack-virtual version

v3.14.9

### TypeScript version

v6.0.3

### Additional context

_No response_

### Terms & Code of Conduct

dikshit-n:
## Summary

Fixes `scrollToIndex(last, { align: 'end' })` so it scrolls to the virtual max offset (content end) rather than the raw DOM max scroll offset (content + paddingEnd) when `paddingEnd > 0`.

## Problem

When `paddingEnd` is set on a virtualizer, `getOffsetForIndex(last, 'end')` returned `getMaxScrollOffset()` = `scrollHeight - clientHeight`. Since `scrollHeight` includes `paddingEnd`, the returned offset overshot the rendered end of the last item by exactly `paddingEnd` pixels. This made `scrollToIndex(count - 1, { align: 'end' })` scroll past the last item.

## Solution

Subtract `paddingEnd` from `getMaxScrollOffset()`:

```typescript
if (align === 'end' && index === this.options.count - 1) {
  return [
    Math.max(this.getMaxScrollOffset() - this.options.paddingEnd, 0),
    align,
  ] as const
}
```

This gives `(content - clientHeight)` — the correct virtual max offset that keeps the last item flush with the bottom of the viewport.

## Why getMaxScrollOffset() and not getTotalSize()?

Using `getMaxScrollOffset()` directly (rather than `getTotalSize() - paddingEnd - getSize()`) preserves the lane-max behavior added in [#1105](https://github.com/TanStack/virtual/pull/1105) ([#1001](https://github.com/TanStack/virtual/issues/1001)). In multi-lane layouts where the last item lives in a shorter lane, `getMaxScrollOffset()` still absorbs DOM extras (borders, padding, unmeasured dynamic items) that aren't in our measurements. Targeting `item.end` would regress #1001 by scrolling the last item above the viewport top.

## Changes

- **src/index.ts**: Subtract `paddingEnd` from `getMaxScrollOffset()` for last-item end alignment
- **tests/index.test.ts**: 3 regression tests for #1257 (paddingEnd overshoot) and #1263 (lane-max preservation)
- **changeset**: patch for `@tanstack/virtual-core`

## Testing

All new tests pass. Existing tests (including the #1258 clamped-growth suite) are unaffected.

Closes TANStack/virtual#1263
Closes TANStack/virtual#1257

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed end-aligned scrolling to the last item when bottom padding is present.
  - The final item now aligns flush with the viewport without overscrolling.
  - Preserved correct behavior for multi-lane layouts and cases without bottom padding.


piecyk (changes requested):
Thanks, this is the right shape: keeping the DOM max from #1105 and only subtracting paddingEnd.

Two things before merge:

- The "#1263: shorter lane" test fails on this branch. Its mock has `scrollHeight: 200` with `clientHeight: 200`, so `getMaxScrollOffset()` is 0, while the comment and the assertion assume a 400px lane-max. Set `scrollHeight: 400` and it passes. Please run the full suite locally before pushing.
- Run prettier on the test file and rebase onto main.

I'll close #1263 in favour of this one.

PR #1284, opened 2026-09-13
fix(core): refresh offset in scroll-end fallback

### LINKED ISSUE #1283: Debounced scroll-end fallback can restore a stale offset after measurement adjustments
### Describe the bug

With `useScrollendEvent: false`, the debounced fallback in `observeOffset` reports the offset captured by the last browser `scroll` event instead of re-reading the scroll element when the debounce fires:

```ts
() => cb(offset, false)
```

If Virtualizer applies item-measurement scroll adjustments after that last event and those programmatic writes do not emit another `scroll` event, the DOM offset advances but the captured `offset` does not. After 150 ms, the fallback resets `scrollOffset` to the stale value. A subsequent measurement adjustment then writes from that stale base and visibly snaps the viewport.

This is not specific to prepending, but a keyed-prepend fixture makes it deterministic enough to measure: every failing run loses exactly **391 px**, the size of the measurement adjustments between the final browser scroll event and the fallback.

Changing the fallback to re-read at callback time fixes it:

```ts
() => cb(readOffset(element), false)
```

Using native `scrollend` (`useScrollendEvent: true`) also avoids the bug on supporting browsers.

### Minimal reproducible example

Public fixture at exact commit:

https://github.com/rubinsh/matrix-bot-web-client/blob/ec7e2b59a20b451505f1562feb18103fced40b01/e2e/timeline-window.spec.ts#L67-L82

The page/virtualizer fixture is in the adjacent `e2e/fixtures/timeline-window.*` files.

### Steps to reproduce

```bash
git clone https://github.com/rubinsh/matrix-bot-web-client.git
cd matrix-bot-web-client
git checkout ec7e2b59a20b451505f1562feb18103fced40b01
pnpm install --frozen-lockfile
pnpm exec vite --host 127.0.0.1 --port 5197

# In another shell:
E2E_BASE_URL=http://127.0.0.1:5197 \
  pnpm exec playwright test e2e/timeline-window.spec.ts \
  --project=mobile --repeat-each=10
```

The failing assertion compares the same keyed row's viewport-relative top before and after prepend.

My independent run at the exact commit produced 3/10 failures, all:

```text
Expected: -42.75
Received: 348.25
Difference: 391
```

Another independent reviewer reproduced 4/10 at the same head. With only `useScrollendEvent: true`, they got 30/30 passes. With native scrollend plus the application's compensating prepend loop deleted, they got 30/30 mobile and 15/15 desktop.

An instrumented failing run showed:

1. one real scroll event records DOM/core offset `37867`;
2. ResizeObserver item measurements adjust DOM/core together to `38258` without another scroll event;
3. the 150 ms fallback calls back with captured `37867`, silently making core state stale by 391 px;
4. the next item measurement writes `stale + delta` to the DOM;
5. keyed prepend anchoring is then correct relative to the wrong baseline, leaving the row 391 px off-screen.

### Expected behavior

The debounced fallback should report the scroll element's current offset when it declares scrolling finished, including any Virtualizer measurement adjustments made since the last browser scroll event.

Sometimes (3/10 and 4/10 in independent repeated runs; timing-dependent).

### Platform

- Linux
- Playwright bundled Chromium 1217
- The stale fallback also affects browsers without native `scrollend`, including older WebKit.

### tanstack-virtual version

`@tanstack/virtual-core` 3.17.8 (resolved by `@tanstack/vue-virtual` 3.13.36)

The same line is currently present on `main`.

### TypeScript version

5.9.3

### Additional context

The current `supportsScrollend` feature check and `useScrollendEvent` option make this primarily a fallback-path bug. A one-line current-offset read fixes the fixture without any application-level anchoring workaround.

- [x] I understand that if my bug cannot be reliably reproduced in a debuggable environment, it may be closed.

rubinsh:
## Summary

The debounced fallback for browsers/configurations without native `scrollend` currently replays the offset captured by the last browser `scroll` event. Virtualizer may make measurement-driven programmatic scroll adjustments during the debounce window without receiving another scroll event, leaving that captured value stale.

Read the scroll target's current offset when the fallback fires instead. This keeps the core's tracked offset aligned with the DOM and prevents the next measurement adjustment from snapping the viewport back to a stale base.

Fixes #1283.

## Verification

- Added a focused fake-timer regression covering a programmatic offset change without a second scroll event.
- `pnpm --filter @tanstack/virtual-core test:lib -- --run` — 165/165 passed.
- `pnpm --filter @tanstack/virtual-core test:types`
- `pnpm --filter @tanstack/virtual-core test:eslint`
- `pnpm --filter @tanstack/virtual-core build`
- Application-level repeated keyed-prepend fixture: failure reproduces at the released code (3/10 locally, constant 391 px) and is eliminated by this current-offset read.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed virtual scrolling to report the latest scroll position after programmatic scroll changes.
  - Prevented delayed scroll handling from overwriting newer measurement adjustments with stale position data.

- **Tests**
  - Added coverage for delayed scroll-position updates after programmatic changes.


piecyk:
Thanks @rubinsh

